Issue #259, opened 2014-08-23
WIP : validation.timezone


caouecs:
"timezone" => "The :attribute must be a valid zone."

Status :

- [ ] Albanian
- [ ] Arabic
- [x] Brazilian
- [x] Bosnian
- [x] Bulgarian
- [ ] Cambodian
- [x] Catalan
- [x] Chinese
- [x] Chinese from Taiwan (zh-TW)
- [ ] Czech
- [ ] Danish
- [x] Dutch
- [ ] Finnish
- [x] French
- [x] German
- [ ] Greek
- [ ] Hebrew
- [ ] Hungarian
- [x] Indonesian
- [x] Italian
- [x] Japanese
- [x] Korean
- [x] Lithunian
- [ ] Macedonian
- [ ] Montenegrin
- [ ] Norwegian Bokmål
- [x] Polish
- [x] Portuguese
- [ ] Romanian
- [x] Russian
- [x] Serbian
- [ ] Slovak
- [x] Spanish
- [x] Swedish
- [x] Thai
- [x] Traditional Chinese (Hong Kong)
- [x] Turkish
- [x] Vietnamese

PR #318, opened 2014-12-19
translated boolean and timezone in ja/validation.php

takahashiyuya:
translated validation.boolean and validation.timezone #259 
translated reminders.reset #258


takahashiyuya:
I sent a pull request. Please check #326.

PR #332, opened 2015-02-05
Update validation.php - timezone and boolean

stojankukrika:


caouecs:
Thanks but can you do same update in laravel5 branch ?

PR #339, opened 2015-02-07
update reminders.php. The password is required 'at least'

zhuzhichao:
The password is required 'at least', but here 'at least' isn't translated.


overtrue:
@zhuzhichao is right.

https://github.com/laravel/framework/blob/5.0/src/Illuminate/Auth/Passwords/PasswordBroker.php#L221

:+1:

overtrue:
@zhuzhichao 建议修改文字 “必须至少” 为 “至少”

caouecs:
Thanks, and it's good for laravel5 branch ?

overtrue:
@caouecs yes!

PR #353, opened 2015-03-06
Spanish-friendly transaltion

nikoskip:
I think "Tu" is more friendly than "Su", it sounds too formal.


caouecs:
Hello,
thank you but can you do same update in laravel5 branch ?

PR #366, opened 2015-04-19
Update validation.php

chotikarn:
Translate new string,
Remove unnecessary period,
Correct "กิโลไบร์ท" to "กิโลไบต์".


caouecs:
Thanks

PR #370, opened 2015-05-21
Add 'string' rule and reposition 'timezone'.

overtrue:
- add `string` rule, FYI: https://github.com/laravel/laravel/commit/11a8db89827072a18b5b8c0b4bf98b8ae9008f43#commitcomment-11300792

- reposition `timezone`


caouecs:
Problem with position of timezone, because the alphabet is : ABCDEFGHIJKLMNOPQRS**T**UVWXYZ	

so thanks for `string`, but it's not good for `timezone`;)

overtrue:
@caouecs Yes, I agree and modified it.

caouecs:
thanks

PR #374, opened 2015-06-08
added some common attributes to validation.php

blueberry-media:
i noticed that the fr/validation.php also has these, and they are quite useful...


caouecs:
Thank you

PR #378, opened 2015-06-15
SQ: Translated all remaining strings and improved/fixed some existing ones

shefi:
Translated *all* remaining strings and *improved/fixed* some existing ones in the **Albanian** language files.


caouecs:
Thanks

PR #385, opened 2015-07-03
Italian: Fix format and added "throttle" support

billmn:


it/pagination.php
```diff
@@ -14,7 +14,6 @@
     */
 
     'previous' => '&laquo; Precedente',
-
     'next'     => 'Successivo &raquo;',
 
-];
+];
```
GrahamCampbell: cs

it/passwords.php
```diff
@@ -13,14 +13,11 @@
...
-
-];
+    'password' => 'Le password devono essere di almeno 6 caratteri e devono coincidere.',
+    'reset'    => 'La password è stata reimpostata!',
+    'sent'     => 'Promemoria della password inviato!',
+    'throttle' => 'Troppi tentativi di accesso. Riprova tra :seconds secondi.',
+    'token'    => 'Questo token per la reimpostazione della password non è valido.',
+    'user'     => 'Non esiste un utente associato a questo indirizzo e-mail.',
+
+];
```
GrahamCampbell: cs

it/validation.php
```diff
@@ -104,4 +105,4 @@
 
     'attributes' => [],
 
-];
+];
```
GrahamCampbell: cs

billmn:
Fixed

it/validation.php
```diff
@@ -34,8 +35,8 @@
     "digits"           => ":attribute deve essere di :digits cifre.",
     "digits_between"   => ":attribute deve essere tra :min e :max cifre.",
     "email"            => ":attribute non è valido.",
-    "exists"           => ":attribute selezionato/a non è valido.",
     "filled"           => ":attribute è richiesto.",
+    "exists"           => ":attribute selezionato/a non è valido.",
```
caouecs: why ?

caouecs:
I accept but I fix the error of order in validation

it/validation.php
billmn: Because Laravel have a different order. 

https://github.com/laravel/laravel/blob/master/resources/lang/en/validation.php

caouecs: I think it's a problem on Laravel version

billmn:
In any case we have to change labels again 

https://github.com/laravel/laravel/commit/ef958716c37fcdf0ea80bb89ba0c93394ae0b157

billmn:
Oh I don't see that you have already updated the package, sorry

PR #389, opened 2015-07-06
Fixed format like original Laravel files

billmn:
Compared to original Laravel files, there are missing blank lines, an example here: https://github.com/laravel/laravel/blob/master/resources/lang/en/auth.php


caouecs:
Hello,
it's critical ?

billmn:
Hi, no it's not critical ... it's just a different format than the original files.
I think may be helpful to have them lined up, that's all ;)

caouecs:
We need to update validation files too ;)

so, for the moment, I close this pull request

billmn:
No problem ;)

PR #398, opened 2015-07-13
Updated the danish (da) translations

KHIT93:
I have updated the translations of the strings into danish and corrected a few strings that were not translated properly.


caouecs:
Tak

PR #405, opened 2015-07-28
Update auth.php

ersin-demirtas:
English to Turkish translation for auth.php file.

PR #410, opened 2015-08-04
Update Auth.php

ahmetalan:


caouecs:
see #405

ahmetalan:
#405 not a true translate.

ahmetalan:
both are correct not problem.

PR #414, opened 2015-08-19
pt_BR - Better translation for required field

nerijunior:
A simple version of the string on required and filled fields.


vinicius73:
+1

caouecs:
Obrigado

PR #421, opened 2015-08-28
Fixed Italian file format and minor improvements

billmn:
- Fixed indentation (like original Laravel file)
- Used ' for array keys but not for values ( to allow apostrophe )
- Fixed some translations (max, min, size, string and url)


caouecs:
Grazie

PR #427, opened 2015-09-16
Fix dutch 'max' translations

boyd91:
The max translations stated ':attribute must be less than :max'. This is incorrect as it should be 'can't be more than :max'.


caouecs:
Thanks

PR #428, opened 2015-09-17
Translate auth.php

ehel:
I translated into Romanian auth.php .


caouecs:
Oups !! Sorry, conflict with this pull request #429

PR #429, opened 2015-09-19
Romanian translation update

avataru:
Most are smaller fixes and improvements but there were also some bigger mistakes such as kilobyte being translated as kilobit instead of [kilooctet](https://en.wikipedia.org/wiki/Octet_(computing)) and missing translation. 

Also, I used the current English language file found in the Laravel repo as a base for formatting and item order.


caouecs:
Thanks

PR #443, opened 2015-10-14
Added Translation to hr/auth.php

Trveller:
This correction added a correct Croatian translation
instead of the english ones that were in the auth.php script.
Users of this code will therefore not have to change/modify
these manually as they are already correct.
This prevents from showing unwanted translations and
enhances the user experience.


caouecs:
Hello,
thanks but can you delete vagrant.xml ?

Trveller:
Hi.

This should be it.  :)

caouecs:
Thanks

PR #456, opened 2015-11-01
Clear translation of 'characters'.

bahjaat:
Clear translation of 'characters'. Where 'karakters' is used before, 'karakters' should be used here too in stead of 'characters'.


caouecs:
Thanks

PR #471, opened 2015-11-27
Updating .gitignore file

arcanedev-maroc:


.gitignore
```diff
@@ -1,3 +1,5 @@
 _site
+/.idea/
 /vendor/
 .DS_Store
+composer.lock
```
caouecs: Why composer.lock ?

arcanedev-maroc: The real question is: why keep it ?

caouecs:
Thanks

PR #477, opened 2015-12-01
Double arrow align and enabled PHP-CS-Fixer Cache

billmn:
Related to this issue: https://github.com/caouecs/Laravel-lang/issues/460


caouecs:
Thanks

PR #480, opened 2015-12-10
Changed confusing Dutch phrasing for required_unless

boukeversteegh:
The Dutch phrase is confusing, it will say:
``admin password is verplicht tenzij rol voorkomt in admin, moderator``
(``admin password is required unless role occurs in admin, moderator``)

Whereas this makes more sense:
``admin password is verplicht tenzij rol gelijk is aan admin, moderator``
(``admin password is required unless role is equal to admin, moderator``)


caouecs:
Thanks

boukeversteegh:
:+1:

PR #485, opened 2016-01-03
Readme: language -> locale

JeroenNoten:
The readme is wrong about the name of the field in `app.php` that we need to edit.


caouecs:
thanks

PR #489, opened 2016-01-10
Add the Galician language translation (gl ISO 639-1)

amieiro:
I have added the 4 translated files for the Galician language (gl ISO 639-1):

auth.php
pagination.php
passwords.php
validation.php


caouecs:
Thank you

PR #510, opened 2016-03-06
fix a typo and improve the translation

djug:
fix a typo in the first sentence (there was an extra 'ت' in 'متطابقة'

in the second sentence we should use "ثانية" instead of "ثواني" since "ثواني" could be used only with numbers between 3 and 10


caouecs:
Thanks

PR #519, opened 2016-03-10
Other Dutch translations

JurgenNL:


caouecs:
Thanks.

Can you translate "distinct" line ?

PR #526, opened 2016-03-19
Updating the folder structure & todo generator script

arcanedev-maroc:
This PR is:
  * to seperate all translations to a dedicated folder (src) to avoid the conflicts with other unrelated folders (like script).
  * Refactoring and opmitizing the todo generator script


caouecs:
If you change directory of translated files, you break compatibility with others packages. I'm ok for en files in script directory.

arcanedev-maroc:
Yes @caouecs, i agree it's gonna have a break compatibility for some packages but it's gonna be once and for a good reason.

We are going to have one folder that contains all the translations (`src` like any php packages), so the other folders like `script` does not mixed up with them.

This way if we scan the `src` folder, we are sure we're going to have only the `languages` and their translations.

And also, we're gonna to have more flexibility if we decide to create a new folder(s) that is not related to translations (like `tests` for examples).

In my opinion, this is a good folder structure and maintainable.

caouecs:
Yes, it's a good idea. But for the moment, I prefer to do step by step. So I'm ok for en files in script directory.

caouecs:
And for todo generator, I prefer an another pull request

arcanedev-maroc:
Hi @caouecs, i've created a new PR for the TODO.md generator script (optimized).

caouecs:
In my todo for this project, I have two important lines :

* Rename the project name for Composer ( destroy laravel-4 )
* update directories

The best is to do a version 3. I will try this week-end.

arcanedev-maroc:
ça marche, merci pour ta réactivité :+1: .

PR #529, opened 2016-03-22
Fix indent

juukie:
Now equal to the [original file](https://raw.githubusercontent.com/laravel/laravel/master/resources/lang/en/validation.php).

### DIFF
--- a/nl/validation.php
+++ b/nl/validation.php
@@ -91,7 +91,7 @@ return [
     |
     */
 
-    'custom'               => [
+    'custom' => [
         'attribute-name' => [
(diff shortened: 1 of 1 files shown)


caouecs:
We don't respect the original file, we have our own phpcs ;)

juukie:
But the [en file](https://github.com/caouecs/Laravel-lang/blob/master/en/validation.php) stays original?

caouecs:
it's my fault

juukie:
Ok. I see you've updated it.

PR #545, opened 2016-04-20
Changes in two aattributes

Kikerrobles:
Sorry, when I was writing commit description somethig was wrong.

In Spain we always use two surnames, so insted of "apellido" we use "apellidos"
In Spain we never use "celular" as mobile (it's used in latin countries), we always use "Móvil"


caouecs:
Gracias

PR #548, opened 2016-05-07
[fa] Complete translations and fix u+200c (half-space) character.

imami:
Translate Some parts that remain untranslated.
Also use u+200c unicode character (Also knowne as Half-Space) in some previous translations.


caouecs:
Thanks

caouecs:
Can you translate *present* in this file ?

imami:
Done :)

PR #554, opened 2016-05-12
update Georgian language pack

Landish:


src/ka/validation.php
```diff
@@ -92,7 +89,7 @@
     |
     */
 
-    'custom'               => [
+    'custom' => [
```
caouecs: Problem of style

src/ka/validation.php
```diff
@@ -109,8 +106,6 @@
     |
     */
 
-    'attributes'           => [
-        //
-    ],
+    'attributes' => [],
```
caouecs: Don't change it

src/ka/validation.php
```diff
@@ -109,8 +106,6 @@
...
 
-];
+];
```
caouecs: Need new line at end

src/ka/validation.php
Landish: Laravel's core validation.php file has been changed and for some reason I changed it as well. https://github.com/laravel/laravel/blob/master/resources/lang/en/validation.php#L112

caouecs:
Thanks

PR #555, opened 2016-05-16
hebrew translation of validation dimensions

boynet:
I think we can create a list of translator in each language that will be pinged when new translation is added, I would love to get pinged for hebrew translation


caouecs:
Thanks

caouecs:
When I'm not sure with a translation, I ping people. but now, you're my reference for hebrew ;)

PR #562, opened 2016-06-27
Updated validation.php for Lithuanian language

fosron:
Updated validation.php for Lithuanian language with validation.dimensions translation.


caouecs:
Thanks

PR #564, opened 2016-06-29
[de] Added translations

tobysommer:
[de] Added translations for "dimensions", "distinct", "in_array" and "present"


caouecs:
Danke

PR #576, opened 2016-07-26
Fixes to the pt-BR translation

fernandobandeira:
These translations were pretty weird, i guess that they were copied from the pt files or made with google translator, as there's a lot of sentences that doesn't make sense.

Also tried to use the same words in the sentences so the messages don't look too different.

Each change is specified on the according commit, there's a lot of changes also fixed some typos in words.


caouecs:
Obrigado

PR #599, opened 2016-09-26
Translated comments for auth.php file of ru

Djasir:
Solution for #598 Not translated auth.php file comments of ru


caouecs:
Thanks

PR #609, opened 2016-10-16
more common dutch

arnoudhgz:
'Gefaalde' is not a common word for failed in internet/website language for Dutch 'mislukte' is better


caouecs:
Thank you

PR #616, opened 2016-10-26
[de] Fixed german translation

plxx:
Just a fix for some inconsistencies in the german translation.

+ php-cs-fixer


caouecs:
Danke

caouecs:
Can you update *uploaded*, please ?

PR #618, opened 2016-10-31
[ca] Updated "uploaded" validation message in catalan

rsancle:
Translated some strings from english to catalan.


caouecs:
Gràcies

PR #619, opened 2016-10-31
[pt-BR] Consistency

fernandobandeira:
Removing the word "indicada(o)" and "especificada(o)" replacing it with "informada(o)" since we use it more in our language and this will keep the messages consistent.

Also forgot to add my name to the sources.md, so i added it...


caouecs:
Obrigado

PR #625, opened 2016-11-12
Grammar mistake in auth.php

Maxim-Mazurok:
"Вовремя" means "in time", here should be used "во время" which means "during"


caouecs:
Thank you

PR #637, opened 2016-12-02
[fi] Added new Finnish translations

Kehet:
Replaced Finnish form validation placeholders with translations


caouecs:
A problem of copy-paste ?

PR #639, opened 2016-12-09
Fix 'min' validation rule for French

shadoWalker89:
The existing translation is saying that the value should be bigger then `:min`

I added the part when it says that it could be equal to `:min`


caouecs:
merci

PR #640, opened 2016-12-11
Fix 'min', 'max' and 'size' validation rules for Arabic

shadoWalker89:
Existing translations for `min` and `max` omit the "or equal" part

For the `size` rules current translations are saying that it should be bigger then `:size` instead of equal to `:size`


caouecs:
Thank you

PR #642, opened 2016-12-14
Added Language Hindi

mramitict:
The most widely spoken language of northern India, with over 200 million speakers; one of the official languages of India. It is an Indic language derived from Sanskrit and is written in the Devanagari script.

I have applied the translation for auth, pagination, passwords and validation files.


caouecs:
Thanks

PR #665, opened 2017-02-22
edit "password" translation

atefBB:
"pass" (see https://en.oxforddictionaries.com/definition/pass) in arabic is «مرّ» ("Marra", see http://www.wordreference.com/enar/pass).


caouecs:
Thank you

PR #675, opened 2017-03-26
[fa] rewrite sentences

mdaliyan:
Just some rewrites to make sentences suitable and easy to understand for common users who don't have any knowledge about programming.

PS: My first language is Farsi.


caouecs:
Hello,
thank you, but not conflict with the [purpose of Anisi](https://github.com/caouecs/Laravel-lang/pull/669) ?

mdaliyan:
I didn't see that coming. sorry for that.

caouecs:
Hello @mdaliyan,
it's not a problem, but, we have a problem in validation.php with attributes.password_confirmation.

I accept your version or his version ??

mdaliyan:
Well, Apart from the fact that I made a lot of change on the file, the only conflict between these 2 versions is "**password_confirmation**" key.

The exact translation for "confirmation" is "تاییدیه"... but this word is not used in this case. We commonly use "تکرار رمز عبور" that is translated to "repeat password". The most common use case of "تکرار رمز عبور" phrase is in ATMs. 

As I said before, these translations are edited to be easy to understand for common users.

May you please tell your opinion @Anisi ?

Anisi:
I agree with your opinion @mdaliyan  about that "تکرار رمز عبور" is more common for users.

caouecs:
Thank you

mdaliyan:
That's right @Anisi . using "پست الکترونیک" bothers me a lot too.

PR #678, opened 2017-03-29
Fix Dutch 'underscore' translation

jeroennoten:
In Dutch, 'liggend streepje' is a dash, see https://nl.wikipedia.org/wiki/Liggend_streepje_(leesteken).
I think it is common to just use 'underscore' in Dutch (https://nl.wikipedia.org/wiki/Underscore).


caouecs:
thank you

PR #685, opened 2017-05-10
Fix Korean translation

flrnrud7:
- Translate untranslated items
- Fix incorrect translation and grammar
- Remove unnecessary words


caouecs:
Thank you

PR #687, opened 2017-05-12
Added IPv4 and IPv6 validation messages for all languages

billmn:


dersvenhesse:
Wouldn't it be feasible to use the already translated strings for `ip` in every case and just replace `IP` with `IPv4` and `IPv6`?

billmn:
@dersvenhesse Not in every case, for example take a look here: https://github.com/caouecs/Laravel-lang/blob/master/src/be/validation.php#L50

dersvenhesse:
I assume `IPv6-адрасам` would work since it translates to `IPv6-address`. But I get your point, it might not be the most suitable way in all cases.

caouecs:
Thank you

PR #690, opened 2017-05-15
Translated IPv4/6 messages to German

dersvenhesse:
Translated the new messages for IPv4/6 validation to German.


caouecs:
Danke

PR #717, opened 2017-06-07
Change :attribute to start with uppercase letter

peterangelov:
Applies if :attribute is at the beginning of a sentence.
+ added translations for IPv4 and IPv6


caouecs:
Thank you

Issue #700, opened 2017-05-25
Letters capitalized


caouecs:
Since Laravel 5.2, we can define letters capitalized. A new branch has been created to add this feature : `capital_letters`.

* [x] ar
* [x] be
* [x] bg
* [ ] bn
* [x] bs
* [ ] ca
* [ ] cs
* [x] cy
* [ ] da
* [ ] de
* [x] el
* [ ] es
* [ ] et
* [ ] fa
* [x] fi
* [x] fr
* [ ] gl
* [x] he
* [ ] hi
* [x] hr
* [x] hu
* [x] id
* [x] is
* [ ] it
* [ ] ja
* [ ] ka
* [ ] kk
* [ ] km
* [ ] ko
* [x] lt
* [x] me
* [x] mk
* [x] mn
* [ ] ms
* [ ] nb
* [ ] ne
* [x] nl : #699 
* [ ] nn
* [ ] pl
* [x] pt
* [x] pt-BR
* [x] ro
* [x] ru
* [ ] sc
* [ ] sk
* [ ] sl
* [ ] sq
* [x] sr
* [ ] sv
* [x] sw
* [ ] tg
* [x] th
* [ ] tk
* [x] tl
* [ ] tr
* [x] uk
* [ ] ur
* [x] vi
* [x] zh-CN
* [x] zh-HK
* [x] zh-TW

overtrue:
zh-* not needed

ctf0:
[ar] is done

caouecs:
Hello all, 
I think it's time to merge this feature in master, for a new version ( 4.x.x )

PR #725, opened 2017-06-21
[NL] Start sentences with capital letters

standaniels:
When starting a sentence with the attribute name, `:attribute` is replaced by `:Attribute` so it starts with a capital letter.


caouecs:
Thank you, but #700

PR #731, opened 2017-07-20
Update validation.php

eXistenZNL:
##What

Fixed a typo (actual a language rule error) in the Dutch translations.

## Why

Because it's against the Dutch language rules to have spaces in compound words. The rules are explained at http://www.spatiegebruik.nl/regels.html :)


caouecs:
Danke

eXistenZNL:
:+1:

PR #733, opened 2017-07-25
Update validation.php

decadence:
Deleted wrong `-` and make link dynamic to latest docs version.
[Wikipedia](https://ru.wikipedia.org/wiki/%D0%9F%D0%BE_%D1%83%D0%BC%D0%BE%D0%BB%D1%87%D0%B0%D0%BD%D0%B8%D1%8E)


caouecs:
Thank you

PR #746, opened 2017-08-27
[de] Additional attributes

kitbs:
Added attribute translations for German based on those appearing in the French file.


caouecs:
Danke

PR #750, opened 2017-09-11
Fix 'filled' validation message

idhamperdameian:
There is something changed with original source, this make also need to change with the translation, see it on https://github.com/laravel/laravel/commit/90886732cf6df6d8694287c3f9d92496a4f1c61b#diff-9c04f727a922ad4f105b9ad764ec221d


caouecs:
Thank you

PR #760, opened 2017-10-11
[PT-BR] Update validation.php with text attribute

gabriel-caruso:
Very common word used in forms, for example, `textareas`


caouecs:
Obrigado

PR #764, opened 2017-10-18
Added & updated some translations

alesvaupotic:
Initial translation for some attributes, changed others. Due to complexity of the Slovenian language, we should probably go thru all of these and make some adjustments with real-world examples. I am setting up a new project right now and will work on forms in about 14 days.


caouecs:
Thank you

alesvaupotic:
You're welcome!

PR #766, opened 2017-10-18
Translated two elements to Bulgarian

martinyakimov25:
Translated 'in_array' and 'required_unless' to Bulgarian


caouecs:
Thank you

PR #769, opened 2017-10-30
Update Norwegian validation translation

torkiljohnsen:
* Added a few strings that were not translated.
* Fixed må/skal (must/shall) to be consistent, so we use "må" now everywhere, instead of a mix of the two.


caouecs:
Takk

caouecs:
Takk

PR #775, opened 2017-11-04
Fixed german translation

TheNuclearCat:
The german translation was not really clearifiying, that the combination was not found.
With the current translation it looks like the data is not existing in the database.


TheNuclearCat:
btw. why does style ci failing? I have changed only a couple of words...

caouecs:
Danke

PR #780, opened 2017-11-20
modified description

Mombuyish:
see `Illuminate\Validation\Concerns\ValidatesAttributes.php`

I found some description does not correct.

\pL matches any kind of letter from any language
\pM matches a character intended to be combined with another character (e.g. accents, umlauts, enclosing boxes, etc.)
\pN matches any kind of numeric character in any script
_- matches a single character in the list _- (case sensitive)

``` PHP
    public function validateAlpha($attribute, $value)
    {
        //including Han.
        return is_string($value) && preg_match('/^[\pL\pM]+$/u', $value);
    }

    public function validateAlphaDash($attribute, $value)
    {
        if (! is_string($value) && ! is_numeric($value)) {
            return false;
        }
        //including Han, underline and dashes.
        return preg_match('/^[\pL\pM\pN_-]+$/u', $value) > 0;
    }

    public function validateAlphaNum($attribute, $value)
    {
        if (! is_string($value) && ! is_numeric($value)) {
            return false;
        }
        //including Han.
        return preg_match('/^[\pL\pM\pN]+$/u', $value) > 0;
    }
```
The description does not includes chinese describe and  translate dashes are not correctly.


caouecs:
@Mombuyish thank you for your work

@overtrue What's your opinion ?

PR #797, opened 2018-03-06
şifre değil parola

ugurarici:
"Şifre" actually means "cipher" (as in encryption) in Turkish, "parola" is the correct translation of "password"


caouecs:
Thank you

PR #799, opened 2018-03-13
Update validation.php

ubrt:
Misspelling in 'password_confirmation'. The german term should be 'Passwort-Bestätigung'


caouecs:
Danke

PR #822, opened 2018-05-10
[it] Translate validation.gt, .gte, .lt, .lte, singularize kilobytes

tegola:
English words aren't used in their plural form in in the italian language.


caouecs:
Grazie

PR #823, opened 2018-05-13
[gl] Update the validation file 

amieiro:
Update the new lines in the validation file for the Galician language (gl ISO 639-1)


caouecs:
Thank you

PR #832, opened 2018-05-19
Change isian to bidang

mabuak:
Change "Isian" to "Bidang"


caouecs:
Thank you, but I need help to be sure.

@idhamperdameian ?

idhamperdameian:
@caouecs I think no need to change, it good with last translation

PR #833, opened 2018-05-21
 [id] Update validation for gt, gte, lt, lte and not_regex

idhamperdameian:
I also fix other translation.

Todo file by running script `php script/todo.php`


caouecs:
thank you

PR #846, opened 2018-06-21
[nb] Max-rule for files is wrong

piperone:
If I translate the current Norwegian max-rule for strings back to English, it reads:
```:attribute must not be shorter than :max characters```
... which is the exact opposite of what this validation error text is meant to convey. After the change, it will translate to:
```:attribute must not be greater than :max characters```


caouecs:
thank you

caouecs:
@piperone can you help us and translate gt, gte, lt and lte ?

piperone:
@caouecs sure

PR #847, opened 2018-06-21
Translations added for gt, gte, lt, lte

piperone:
Translations added for gt,gte,lt,lte. Tried to keep it similar to existing translations.
Also changed ```kilobytes``` to ```kilobyte``` (s-endings are not used to indicate plural in Norwegian).


caouecs:
Takk

piperone:
De rien

PR #848, opened 2018-06-25
Fix some german translations

xama5:
This PR includes:

- Not all attribute names can be used with the article 'Das'. Thus, removing it completely.

  - Example: "Der Bild konnte nicht hochgeladen werden." is wrong, it should be "Das Bild konnte nicht hochgeladen werden." 

(I have a mix-up in my commit message. I meant 'Der' instead of 'Das'.)

- Fix comparison (grammar) of 'groß"

- Run PHP Fixer, as required per the contribution guidelines.


xama5:
@caouecs I suggest to remove the CI or fix the contribution guidelines, as they do not seem to like each other.

PR #849, opened 2018-06-25
[de-CH] Add Swiss standard German 

xama5:
The Swiss do not use "ß". This PR creates a new language and replaces ß with 'ss'.

For reference: https://en.wikipedia.org/wiki/%C3%9F#Switzerland_and_Liechtenstein

**Attention**: Although it would be funny, this is NOT swiss-german. This is just the localized version of German for Switzerland, thus de-CH.

### DIFF
--- a/src/de-CH/auth.php
+++ b/src/de-CH/auth.php
@@ -1,0 +1,17 @@
+<?php
+
+return [
+    /*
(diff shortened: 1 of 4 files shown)


caouecs:
Danke

PR #863, opened 2018-07-31
[NL] Use more common word "tekst" for string

megaxlr:
'tekenreeks' is very uncommonly used and will likely confuse users. 'tekst' on the other hand is a more common phrase to refer to text. 

Signed-off-by: MegaXLR <admin@megaxlr.net>


robindirksen1:
+1

caouecs:
dankjewel

PR #880, opened 2018-09-10
[ko] translate json file

jun9358:


jun9358:
Um... maybe misplaced this json file. close and recreate pull request after move json file to /json/ko.json

PR #897, opened 2018-10-03
Fix some translation errors in indonesian

therusetiawan:
I see some indonesian translation is not good enough (some dictions is incorret). This pull request will fix that.


caouecs:
Thank you

PR #911, opened 2018-10-12
[fr] Use "email" everywhere and add missing strings in fr.json

hedii:


caouecs:
A contre coeur, je valide...

PR #916, opened 2018-10-22
Removing double arrow alignments

cyberhicham:
Complying with laravel style in lang files


andrey-helldar:
@cyberhicham, why do you remove positioning values?
Positioning values along a single vertical line makes it easier to read files during development.

PR #925, opened 2018-10-29
feature: add new json strings and update todo

clemir:


caouecs:
thank you

TopFuel:
The string `Sorry, you are forbidden from accessing this page.` was changed from `Sorry, you are forbidden to access this page.` in this commit https://github.com/laravel/framework/commit/bb81cb0dfe2e3b8fbef91690abe459d1ee6d18c5
This change has not yet been released, which should happen soon, so the translation of this string will not work for now.

PR #937, opened 2018-11-05
Update AR

mohamedsabil83:


todo.md
```diff
@@ -1,6 +1,5 @@
 # Todo list
 
-* [ar](#ar)
```
caouecs: You don't use "php script/todo.php"

mohamedsabil83: So, you remove this PR and I make another one after using `php script/todo.php` or next time?

caouecs:
Thank you

PR #949, opened 2018-11-21
Updating greek translation gt,gte,lt,let,regex,uuid

seldimi:
Updating greek translation gt,gte,lt,let,regex,uuid


caouecs:
I fixed the end file, and updated the todo file

PR #950, opened 2018-11-22
[de] Added translation for UUID

andrey-helldar:
The todo list says that a translation is needed for the field `Name` in json.
In German, `Name` is written in the same way as in English, so this entry will be in the todo list.


caouecs:
Duplicate "Name", it's not a problem, and the todo list will be good ;)

andrey-helldar:
Granted, this is not a problem. Just a note)

andrey-helldar:
Bitteschön 👌

PR #963, opened 2018-12-03
[ua|ru] added translations starts_with and date_equals, uk.json

vladshoob:
Added starts_with and date_equals translations to both languages (ru|ua).
Added "verify email address" to uk.json, double checked with what it actually used for in Laravel.
Removed items from todo.


caouecs:
Thank you

PR #976, opened 2019-01-02
update comment to match Taylor's style

Khuthaily:
Taylor makes every line of a block comment 3 characters less than the previous line. This change is a copy-paste from his comments in the English version.


caouecs:
thank you

PR #978, opened 2019-01-03
fix grammar error in predicate of kaan - AR

Khuthaily:
A grammatical rule requires the predicate (khabar) of "kaan" to be in the accusative case (mansoob). Thus, the ending "yaa + noon" for the dual is expected rather than the current "alif + noon."


caouecs:
thank you

PR #979, opened 2019-01-04
[KO] Fixed translation error

JaehyeokChoi:
Typically, "ID 저장" is simply used to store the ID(cookie).
"로그인 상태 유지" is more accurate because Laravel's Remember Me is a feature that keeps logged in.


caouecs:
thank you

PR #992, opened 2019-01-23
Fix: markdown link should not have any space

ankurk91:
Hi,
there should not be any space between
```
[:actionURL] (:actionURL)
```

Markdown parser is not considering it a link when there is a space.


caouecs:
Gracias

PR #993, opened 2019-01-23
[it] added a few missing translations

troccoli:
This PR adds almost all the translations in Italian from the todo file. Three are left:

- _click here to request another_: this is difficult to translate correctly as the "another" translation depends on the gender of what we are talking about.
- _Oh no_ and _Password_ are already translated, although for some reason the `script/todo.php` script did not remove them from `todo.md`


caouecs:
Gracie

PR #995, opened 2019-01-25
urdu translations added

fhdalikhan:


caouecs:
Thank you

Can you do translations on validation.php file ?

PR #1000, opened 2019-02-13
[nl] Add 403 Dutch translation

Epse:
The Dutch json file did not contain a translation for the 403 error message. This fixes it


caouecs:
Dank je

PR #1005, opened 2019-03-01
[pl] Update the password string

pawelmysior:
The default required length for the password is eight digits as of Laravel 5.8.0.


caouecs:
thank you

PR #1011, opened 2019-03-08
[kn] Kannada New Language

raghavcool:
[kn] Kannada New Language Translations Added


raghavcool:
Will Raise new PR with PHPCS fix

PR #1029, opened 2019-03-15
[5.8][VI] Add translation for custom attributes

dinhquochan:


caouecs:
No needed, I will merge laravel58 with master ;)

but thank you

PR #1031, opened 2019-03-18
[ru] Translated "This password reset link will expire in :count m"

andrey-helldar:
[ru] Translated "This password reset link will expire in :count minutes."


caouecs:
thank you

PR #1033, opened 2019-03-20
[en] Fix example json strings

pawelmysior:
I found some lines in the example `en.json` files that are not used in the framework so I removed them. I also found two strings that were missing.


ycs77:
Fix to #1060

PR #1034, opened 2019-03-20
[en] Fix english lang files

pawelmysior:
I thought it is a good idea to keep the example english files exactly like they are in the laravel/laravel repo.


caouecs:
No thank, we prefer our style

PR #1038, opened 2019-03-27
JSON file translated to Kannada Version [kn]

raghavcool:
JSON file translated to Kannada Version [kn] as requested


caouecs:
thank you

PR #1045, opened 2019-04-05
[nl] Fix incorrect Dutch url validation message

verwer:
The validation message said ':attribute has to be **no** valid URL'.

That is incorrect and will be solved by this pull request.


caouecs:
Thank you

PR #1048, opened 2019-04-08
Update lang id

donyahmd:
Add missing language on indonesia language.
kilobita changed to kilobyte, because indonesian prefered to say kilobyte than kilobita.


caouecs:
thank you

donyahmd:
Thank's for merged😁

PR #1050, opened 2019-04-10
My give back for providing such amazing repo :)

harshitpeer:
Just helping out with some hindi translation 👍


caouecs:
Thank you

Issue #1051, opened 2019-04-12
Serbian country code issue


arnyee:
Hello, The Serbian folder nem (sr) is not correct, the correct country code is rs!

caouecs:
Hello,
we respect the iso 639-1 and it's sr

petercoles:
@caouecs

Great package. And right answer on this issue, though for the wrong reasons.

It is true that the ISO 639-1 code for Serbian is sr whereas the ISO 3166 code is rs, which I suspect prompts this issue. However, you also use pt-BR, zc-CN, zh-HK and zh-TW, which aren't ISO 639-1 codes.

I'd suggest the standard that actually matters here is the PHP locale codes. For that the 14 or so flavours of Serbian start "sr" as you have it, so won't break anybody app if they include it.

caouecs:
iso 3166 is for country, not language

PR #1052, opened 2019-04-13
#1051: Serbian country code issue

lehone-hp:
Changed Serbia folder name from sr to rs


caouecs:
Hello,
we respect the ISO 639-1 ( https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes ) and it's sr

lehone-hp:
oh okay, closing the pr

PR #1054, opened 2019-04-13
[th] Laravel58 update Thai  translation

ninearif:
- Add missing Thai translation string.
- remove 'full stop' from Thai translation string (we don't use full stop in Thai).
- improve some previous translated string.


caouecs:
The laravel5.8 version is in laravel58 branch for the moment.

sorry

PR #1064, opened 2019-05-08
Ukrainian translations update

Somethingideally:
Updated few messages in ukrainian translations


caouecs:
Thank you, and sorry for the lag

PR #1074, opened 2019-06-17
Add missing Slovene translations

morpheus7CS:
I have added the missing translations for Slovene files as well as the main sl.json file.


caouecs:
thank you

PR #1078, opened 2019-07-16
Modify Lithuanian pagination translation

ekateiva:
According to Lithuanian language rules word "sekantis" can not be used in pagination context.

References:
- http://www.vlkk.lt/konsultacijos/2710-sekantis
- http://zemynosgimnazija.lt/kalbekime_taisyklingai/sekantis_kitas.pdf


caouecs:
Thank you

PR #1093, opened 2019-08-09
Add missing translations (Done by Cover Genius)

specialtactics:
Hi there

We at Cover Genius have found this package very useful, and we've translated various missing translations in the languages we use, plus added Filipino (fil).

Thanks


caouecs:
Thank you very much for your work

caouecs:
I added validation.ends_with, and [todo.md](https://github.com/caouecs/Laravel-lang/blob/master/todo.md)

PR #1096, opened 2019-08-16
Update the Galician language translation (gl ISO 639-1) 

amieiro:
I have updated 3 translated files for the Galician language (gl ISO 639-1):

auth.php
passwords.php
validation.php
I have added the JSON translation file for the Galician language (gl ISO 639-1).


caouecs:
Hello,
gracias pero I prefer gl.json ;)

PR #1101, opened 2019-08-20
Fixed output password length from 6 to 8 characters

andrey-helldar:
Actually for the BE, BG, BS, CY, EL, RO and RU languages.
Others did not look. Sorry 😇


caouecs:
thank you

PR #1113, opened 2019-09-29
Add missing translations for de-CH

chefe:
* validation : date_equals
* validation : ends_with
* validation : starts_with
* validation : uuid


caouecs:
Danke

PR #1114, opened 2019-10-01
Update indonesian language

lakuapik:
I recently open the `id.json` file and found that its not updated yet. I update the `id.json` file according to `todo.md#id` file.

Terima kasih.


caouecs:
thank you

PR #1115, opened 2019-10-01
[ar] Update validation.not_in

shadoWalker89:
The old translation was not clear this what is says
`:attribute exists`
I changed to the equivalent of the original laravel translation which says
`The selected :attribute is invalid.`


caouecs:
thank you

PR #1134, opened 2019-10-18
 [ro] updated validation; changed auth string; added translated json

firewizard:
* updated validation
* changed auth string
* added translated json


caouecs:
thank you

PR #1146, opened 2019-10-28
[es] The password field is not used anymore on Laravel 6

JuanDMeGon:
The password field in the passwords file is not used anymore on Laravel 6. Instead, it uses the password field in the validation file.

### DIFF
--- a/src/es/passwords.php
+++ b/src/es/passwords.php
@@ -12,7 +12,6 @@ return [
     |
     */
 
-    'password' => 'Las contraseñas deben coincidir y contener al menos 8 caracteres',
     'reset'    => '¡Tu contraseña ha sido restablecida!',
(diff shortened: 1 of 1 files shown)


caouecs:
Gracias, but the project must be compatible with Laravel 5

arcanedev-maroc:
@caouecs You need to tag a version compatible only for laravel 6 because it cause an issue for packages like https://github.com/ARCANEDEV/LaravelLang.

It has a translation checker that fails by telling me that `passwords.password` is missing for `en` locale.

Any suggestion ? Thanks

arcanedev-maroc:
ping @caouecs

caouecs:
I can do a branch for Laravel 5 and master for Laravel 6

arcanedev-maroc:
OK @caouecs and you should think also for a specific tag for Laravel 6 only (major version)

PR #1147, opened 2019-10-28
[es] Adding a commonly used attribute (terms)

JuanDMeGon:
"terms" is commonly used for referencing terms of service and similars.


caouecs:
Gracias

PR #1148, opened 2019-10-28
[es] Sorting attributes for consistency and ease

JuanDMeGon:
All the validation file is sorted alphabetically, except for the attributes array. Now, it is solved for consistency and referencing ease.


caouecs:
Gracias

PR #1150, opened 2019-10-31
[ko] Update translation

tbear18:
Update translation of "ends_with" in Validiation.php
and updated the translation of the "ko.json" file.

If there is something wrong, I hope you send me a message.


caouecs:
Thank you

PR #1152, opened 2019-11-01
[zh-*] Optimize string starts_with, ends_with

ycs77:
Optimize the string to make `starts_with` and `ends_with` symmetrical.


caouecs:
Thank you

PR #1158, opened 2019-11-07
[si] Remove empty lines before the PHP opening tags

tvbeek:
While using https://github.com/ARCANEDEV/LaravelLang I did see white spaces in my console. During the debugging I found out that it was because in the si language files there where 2 files with an empty line before the opening tag. This PR removes them.


caouecs:
thank you

PR #1162, opened 2019-11-17
[fr] Added missing translations

WhereIsLucas:
I added the two missing translations in french, and executed php-cs-fixer (this changed a lot of files)


caouecs:
Merci

PR #1167, opened 2019-11-26
[pt-BR] translating throttled line

Higorhicaro:
Only edited src/pt-BR/passwords.php the other changes was PHP-cs-fixer.


caouecs:
Thank you, but just do a pull request for your translation

PR #1169, opened 2019-11-29
[pt-BR] Updated json strings and added translation for ends_with and password in validation.

TopFuel:
Seems like the json string: `We won't ask for your password again for a few hours.` got removed in these commits: https://github.com/laravel/ui/commit/d72626733ead7b8482358f9a27e46ac32da6561a and https://github.com/laravel/ui/commit/23886c7796b2cc8f5595761bd6a77b41b61fedae


caouecs:
Obrigado

PR #1174, opened 2019-12-11
[da] New Language

telkins:
Just a quick/simple replacement using Google Translate.


caouecs:
Thanks but are you sure ? Google Translate ?

telkins:
> Thanks but are you sure ? Google Translate ?

@caouecs Not 100%.  I'm not Danish, but I live in Denmark and speak a fair amount.  I looked through it and it looks fine...but I'm not an expert by any stretch of the imagination.  :-)

telkins:
@caouecs Why did you close this?  If you fear that the Danish isn't perfect, I don't think you should worry.  Like I said, I used Google Translate, but that was to get them all done quickly.  While I'm not Danish, I have spoken Danish since first coming here in 1994.  It's not like I'm rolling the dice.

Anyway, please let me know what your concerns are.  I can run the list by any of my native-Danish colleagues, friends, neighbors, or kids.  :-)

PR #1186, opened 2020-01-18
[uz-latin] Updated passwords && validation messages in Uzbek 

sayidazim772:


caouecs:
Hello,
why there is a lot of updates ?

caouecs:
Thank you

PR #1193, opened 2020-01-26
Normalize translation rules

xorock:
When using custom validation attributes it's very hard to keep field name in an uniform format. Sometimes it starts with :attribute, another time it's inside or at the end of the message. For example:

```php
'attributes' => [
        'password' => 'hasło',
],
```
with regex and min rules:

```
Format hasło jest nieprawidłowy. #regex, sounds strange in polish, stylistically incorrect
hasło musi mieć przynajmniej 8 znaków. #min, starts with lower letter 
```

With modified messages:

```
Format pola hasło jest nieprawidłowy.
Pole hasło musi mieć przynajmniej 8 znaków.
```


caouecs:
Thank you

PR #1198, opened 2020-01-31
SV: Make date validation messages also work with fields

perifer:
This makes the after and before validation messages work equally well with dates and fields, just as the english variant.


caouecs:
Thank you

PR #1200, opened 2020-02-08
[uz-latin]

trollsk1n:
Updated uz-latin messages


caouecs:
Hello,
thank you for your translations, but don't touch at todo.php

PR #1203, opened 2020-02-09
[sk] Added missing password and validation translation

terrdus:
Added missing lines to passwords.php and validation.php

sk.json - many changes are only because VSCode decided to remove empty space at the end of lines.


caouecs:
Thank you

PR #1204, opened 2020-02-12
Turkish language 

monurakkaya:
I've noticed there are things missing in tr.json and other tr related sources


caouecs:
Thank you

PR #1206, opened 2020-03-01
[de] & [de-CH] Fix typos, add commas, etc.; remove non-dialect inconsistencies

MaximeDelwaulle:
…between both languages 

- [de] and [de-CH] are almost the same language, when written, apart from dialectic preferences for certain words ("grösser" VS "größer"); there were fundamentally different ways of phrasing simple things, which I converged here and there
- "E-Mail-Adresse" https://www.duden.de/rechtschreibung/E_Mail_Adresse
- "Passwortbestätigung" VS "Passwort-Bestätigung" -> in theory, both are possible, but googling for "passwortbestätigung" yields 60x more results
- fix typos all around, add commas
- rephrased sometimes to come closer to the original meaning in [en]
- translated 'throttled', which was still in English in both languages

The vast array of minor file edits comes from running the command "php-cs-fixer fix" specified in CONTRIBUTING.md 🤷‍♂️

✨


caouecs:
Hello,
thank you, we have a problem with "php-cs-fixer fix" ;)

Can you do a new pull request with just translation on de and de-ch ?

PR #1208, opened 2020-03-01
fix typo in passwords file - AR

Khuthaily:
replace hamzat-al-qat' with hamzat-al-wasl in the Arabic word for waiting since this noun is generated from a khomaasy verb


caouecs:
Thank you

PR #1209, opened 2020-03-01
fix typo in validation file - AR

Khuthaily:
replace tilde with the indefinite marker of tanween with fatHah


caouecs:
Thank you

PR #1211, opened 2020-03-06
Fix translation in es.json, translation wasn´t work in mail subcopy

luisprmat:
This translation didn't work after updating to Laravel 7.0.*
error-email


caouecs:
Gracias

PR #1228, opened 2020-03-19
HU translate added

zsoltjanes:


caouecs:
thank you

hajnalbendeguz:
What is the 'short name' for hungarian? I've tried to set the locale to 'hu', 'hu_HU' and 'hu-HU'. I see the translations for what I need in the source, but I can not get it to work.

PR #1230, opened 2020-03-20
[ru] fix login failed message

dwcoaching:
Fixed 'login failed' message

Before it was: 'Account name and password DON'T MATCH [each other]'
Now: 'Wrong account name or password'


caouecs:
Thank you

PR #1240, opened 2020-04-20
[de] Fix grammar mistakes and reduce differences between de and de-CH

limenet:
Thanks for this package!

While using the JSON files I noticed `de` and `de_CH` differed in several instances where the two usually don't differ and there were a few smaller grammatical mistakes in both languages (commas and cases) in addition to inconsistent word choice ("Verifikation" vs. "Verifizierung").

Let me know if you'd like me to make any changes!


caouecs:
Danke

but @andrey-helldar are you okay with this changes ?

andrey-helldar:
@caouecs, yes, all right)

I don't use hard binding.

PR #1243, opened 2020-04-28
[tr] Update Turkish translations

deligoez:
With this PR I updated the Turkish translations with far better and more expressive ones.
I also fixed some inconsistent word choices and grammatical mistakes.


caouecs:
Thank you

PR #1245, opened 2020-04-29
[fa] Improvements regarding to Iran's slang

HosseyNJF:
The Farsi/Persian translation (with Arabic script) is mainly used in Iran, and some terms inside the translation files are too formal to be used inside websites. I edited them to make them match more with the native language of Iranians.


caouecs:
thank you

PR #1251, opened 2020-05-13
[id] Uppercase first word

idhamperdameian:
The first word in validation message need to uppercase letter in the Indonesian language


caouecs:
Thank you

PR #1253, opened 2020-05-16
Added missing custom attributes in Validation php

teofanis:
Edited the validation file to add custom attributes for the Greek-language and moved the validation to the correct folder, as I had previously left it in the root.


caouecs:
Thank you

PR #1256, opened 2020-05-22
[fr] Update missing messages

Max13:
These adds are related to 2 new localizable strings in `laravel/ui` introduced by laravel/ui#105


caouecs:
merci

PR #1261, opened 2020-06-11
[sr-cyrillic] Added

LukaLatkovic:
Added [sr- cyrillic], Renamed [sr] to [sr-latin]

**!** I think that the script for todo cannot read Cyrillic letters, that is why in todo.md it says that sr-cyrillic is not finished but it is.


caouecs:
Thank you

arcanedev-maroc:
@caouecs there is a typo in the filename, it's `sr-cyrillic.json` and not `sr-cyrilic.json`

PR #1265, opened 2020-06-19
[PL] Translate the new phrases

olsza:
Polish should already be completely translated :)


caouecs:
Thank you

olsza:
ah, what a blunderer, I made corrections and made a commit, not from this accountn :/  it was to be my private and not a business one ;p
Next time the translations will be from the prv account :)

PR #1266, opened 2020-06-21
[nb] Improve translations

lasseeee:
Improve language to better convey the English meaning. Fix whitespace.


lasseeee:
The style CI seems to want to add the old way of adding whitespace in arrays.

caouecs:
Thank you, I will fix

PR #1273, opened 2020-07-02
Updated nl validation to be consistent with Laravel 6

Rkallenkoot:
Might be a good idea to do this for all languages


caouecs:
thank you

PR #1287, opened 2020-07-13
validation-inline for FA

mehranabi:
Hello everyone,
Hope you are healthy and well.

 I translated validation-inline.php for Persian (Farsi, FA) language.


caouecs:
Thank you.

PR #1297, opened 2020-07-27
[sq] Translate strings which were not translated yet

uhajzeraj:


caouecs:
Thank you.

Can you do the translation of validation-inline.php ?

uhajzeraj:
Hey, sure!

I'll do so as soon as I get the chance 😃

PR #1303, opened 2020-08-06
Create ne.json

dbasechhetri:
A new ne.json file is created for Nepali language.
Few words are added. More to come.


json/ne.json
```diff
@@ -0,0 +1,40 @@
+
```
caouecs, line 1: Please, translate just official json https://github.com/caouecs/Laravel-lang/blob/master/script/en/en.json

PR #1309, opened 2020-08-17
[tr] Update some inline validations.

ismailatkurt:


caouecs:
Thank you, but can you translate other tags ?

PR #1326, opened 2020-09-10
[ko] Translate validation-inline

SilNex:


caouecs (changes requested):
Thank you, just a problem with tag

SilNex:
sorry click miss

SilNex:
> Thank you, just a problem with tag

can you explain the problem of tag??

src/ko/validation-inline.php
```diff
@@ -13,110 +13,110 @@
...
-    'same'                 => 'The value of this field must match the one from :other.',
+    'not_in'               => '선택된 값은 올바르지 않습니다.',
+    'not_regex'            => '형식이 올바르지 않습니다.',
+    'numeric'              => '숫자여야 합니다.',
+    'password'             => '비밀번호가 잘못되었습니다.',
+    'present'              => '이 필드가 있어야 합니다.',
+    'regex'                => '형식이 올바르지 않습니다.',
+    'required'             => '이 필드는 필수입니다.',
+    'required_if'          => ':other이(가) :value 일 때 이 필드는 필수입니다.',
+    'required_unless'      => ':other이(가) :values에 없다면 :attribute 이 필드는 필수입니다.',
```
caouecs, line 101: the problem is here ( sorry I don't validate the review )

caouecs (changes requested):
Here ;)

src/ko/validation-inline.php
caouecs: No :attribute in validation-inline.php

SilNex: thank you 
i'll fix it

caouecs:
Thank you

PR #1329, opened 2020-09-14
Removed 'amount' from validation.php

madman-81:
Removed the Dutch translation for 'amount' since this is only one possible meaning and a specific one. This translation is 'the amount of money' or 'price'. A better translation for the general form/meaning of 'amount' would be 'hoeveelheid'. Since this word isn't it other language files (checked FR, DE and the English source), I recommend to remove it to avoid confusion.

### DIFF
--- a/src/nl/validation.php
+++ b/src/nl/validation.php
@@ -143,7 +143,6 @@ return [
     'attributes' => [
         'address'               => 'adres',
         'age'                   => 'leeftijd',
-        'amount'                => 'bedrag',
         'available'             => 'beschikbaar',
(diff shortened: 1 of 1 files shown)


caouecs:
In 'attributes', we have some translated words, nothing official.

So we can have it, as you want

madman-81:
I'm not sure if I understand you correctly. My recommendation would be to remove this translation. Although this might be a braking change for some applications, changing it now would also cause unexpected behaviour.

Ps. it might be good if attributes would also have some official set of translations, otherwise we still can't use these language files to translate them.

PR #1331, opened 2020-09-14
[NL] Added Dutch translations for validation-inline.php

madman-81:
Added Dutch translations for validation-inline.php


caouecs:
Dak

PR #1332, opened 2020-09-14
[hu] Fix exists validation message

veencel:
Hi,

Currently, the validation message for the "exists" rule is wrong in hungarian.
The original message is: "The selected :attribute is invalid."
The current translation means: "The :attribute exists."

Correctly, it should be "A kiválasztott :attribute érvénytelen."
"kiválasztott" means selected
"érvénytelen" means invalid


caouecs:
köszönöm

PR #1335, opened 2020-09-17
Added translations for Laravel Jetstream + ru translated

andrey-helldar:
image

It's not a mistake.

Laravel treats these as different keys.
I checked it like this:

// resource
image

// command
image

// result
image


caouecs:
It will be hard to review all :)

PR #1354, opened 2020-09-23
Added information about availability of translations for Laravel Jets…

andrey-helldar:
…tream and Fortify

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Laravel-lang
 
-In this repository, you can find the lang files for the framework PHP, [Laravel 4/5/6/7/8](https://laravel.com).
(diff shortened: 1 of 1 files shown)


caouecs:
Fortify ? I missed an episode :)

andrey-helldar:
Yep)

https://github.com/laravel/fortify

@luisprmat [provided links](provided links to translation keys.) to translation keys:

* [The provided password does not match your current password.](https://github.com/laravel/fortify/blob/1.x/stubs/UpdateUserPassword.php#L27) - Update password
* [The provided password was incorrect.](https://github.com/laravel/fortify/blob/1.x/src/Http/Responses/FailedPasswordConfirmationResponse.php#L19) - Password Confirmation
* [The provided two factor authentication code was invalid.](https://github.com/laravel/fortify/blob/1.x/src/Http/Responses/FailedTwoFactorLoginResponse.php#L18) - Two Factor code validator

luisprmat:
Laravel Fortify also has translation strings and it is used for Laravel Jetstream.
Therefore if we do not translate those strings, parts will remain untranslated when we use laravel jetstream

PR #1371, opened 2020-09-28
[es] Translate 'Pagination Navigation'

luisprmat:


andrey-helldar:
I like the difference in languages:

EN: "Pagination Navigation"
ES: "Navegación por los enlaces de paginación"
RU: "Навигация"

😄

caouecs:
Gracias

luisprmat:
@andrey-helldar Spanish is not a language as cheap (low-cost, short in words) as English or others. So I couldn't find another way to describe this phrase in Spanish in this context without changing its meaning.

andrey-helldar:
> @andrey-helldar Spanish is not a language as cheap (low-cost, short in words) as English or others. So I couldn't find another way to describe this phrase in Spanish in this context without changing its meaning.

I am not criticizing. On the contrary, I say that languages ​​are interesting. Somewhere you can describe it in one word, but somewhere the whole sentence is not enough. And that's great.

PR #1373, opened 2020-09-29
Translate for PL

olsza:
ps. the word "Email" is spelled the same in Polish

ps2. Hope it's ok now, thanks @andrey-helldar

### DIFF
--- a/json/pl.json
+++ b/json/pl.json
@@ -82,6 +82,7 @@
     "Once your account is deleted, all of its resources and data will be permanently deleted. Before deleting your account, please download any data or information that you wish to retain.": "Po usunięciu konta wszystkie jego zasoby i dane zostaną trwale usunięte. Przed usunięciem konta pobierz wszelkie dane albo informacje, które chcesz zachować.",
     "Page Expired": "Strona wygasła",
     "Page Not Found": "Nie znaleziono takiej strony",
+    "Pagination Navigation": "Paginacja",
     "Password": "Hasło",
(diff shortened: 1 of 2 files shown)


andrey-helldar (approved):

andrey-helldar:
@caouecs, allow to merge :)

andrey-helldar:
PS: @olsza, the todo.md file is generated automatically when the `php script/todo.php` command is invoked:
https://github.com/Laravel-Lang/lang/blob/master/CONTRIBUTING.md#update-todo-file

luisprmat:
Reviewing in the script `todo.php` to generate `todo.md` we watch that its compares `en.json` with each language searching strings non-translated.
If a string is in `en.json` but it isn't in the language is marked like `: not present`. But if it's as `en.json` as in the language`.json` it expects than translation is different to `en.json`, otherwise that string is added to *todo list* and the flag for this language never is green. 
Is there a way to fix this issue when the translation from English matches with translation for a specific language?

andrey-helldar:
So far we have not looked for such a way, but I am thinking about it.

caouecs:
Thank you

caouecs:
for the fix, it's will be difficult, or we do a file with exceptions by language that todo.php reads before creating of todo.md

andrey-helldar:
> for the fix, it's will be difficult, or we do a file with exceptions by language that todo.php reads before creating of todo.md

This is a separate topic for conversation. I'll think about it a bit longer and create a discussion.

PR #1380, opened 2020-10-01
docs: update password translation in malay

kamaroolkarim:


src/ms/validation.php
```diff
@@ -92,7 +92,7 @@
     'not_in'               => ':attribute tidak sah.',
     'not_regex'            => 'Format :attribute adalah tidak sah.',
     'numeric'              => ':attribute mesti nombor.',
-    'password'             => 'The password is incorrect.',
+    'password'             => ':attribute tidak sah.',
```
caouecs, line 95: Why do you add :attribute ?

caouecs (changes requested):
Thank you, but a typo

src/ms/validation.php
kamaroolkarim: so no need attribute? let me change to 'Kata laluan tidak sah' then

andrey-helldar: The original text of this line indicates that the password is incorrect. It has no reference to the attribute name.

andrey-helldar (approved):

src/ms/validation.php
kamaroolkarim: @andrey-helldar ok noted. already fixed it

caouecs:
Thank you

PR #1401, opened 2020-10-07
Auto-make todo.md

andrey-helldar:
Recently, I have been seeing a problem with the todo.md file conflict.

I propose a solution to the problem: when creating a PR, developers DO NOT add the todo.md file to it.

When a PR is accepted into the master branch, the "GitHub Action" is launched, which itself generates this file and pushes to master.

All you need to do in this repository is to check if GitHub Actions are enabled.

For example,
image

If the developer adds a modified todo.md file to the PR, then it's okay. There will be two options:
1. If there is no PR conflict, then it can be accepted;
2. If there is a conflict, the developer will have to solve it.

In any case, the file will be automatically updated when a PR is merged or a push to the master branch.

### DIFF
--- a/.github/workflows/make-todo.yml
+++ b/.github/workflows/make-todo.yml
@@ -1,0 +1,42 @@
+name: make-todo
+on:
+    push:
+        branches:
+            - master
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+
+        strategy:
+            fail-fast: true
+
+        name: Make TODO.md
+
+        steps:
+            -   name: Checkout code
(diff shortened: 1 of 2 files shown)


caouecs:
Let's go to test it ;)

caouecs:
It's cool: https://github.com/Laravel-Lang/lang/actions/runs/293979441

but when we don't have change on todo.md, the action is in error: https://github.com/Laravel-Lang/lang/actions/runs/293968391

andrey-helldar:
I'll fix it now :)

PR #1402, opened 2020-10-07
[it] validation-inline

masterix21:


src/it/validation-inline.php
```diff
@@ -13,110 +13,110 @@
...
-    'string'               => 'This must be a string.',
-    'timezone'             => 'This must be a valid zone.',
-    'unique'               => 'This has already been taken.',
-    'uploaded'             => 'This failed to upload.',
-    'url'                  => 'This format is invalid.',
-    'uuid'                 => 'This must be a valid UUID.',
+    'starts_with'          => 'Deve iniziare con uno dei seguenti: :values.',
+    'string'               => 'Deve essere una stringa.',
+    'timezone'             => 'Deve essere una zona valida.',
+    'unique'               => 'E\' già in uso.',
```
caouecs, line 116: Use " to not have \\'       "E' già in uso"

caouecs (changes requested):
Grazie but just a problem

masterix21:
Done. 

If you like, I could be a referent for the Italian language. I'm working on the other translations.

caouecs:
Grazie for the translations and your proposition to be referent ( we come back fast to you )

PR #1406, opened 2020-10-08
fixed 1377 - ignores the additional information

olsza:
I deleted the old branch and created a new one after updating original:master

### DIFF
--- a/script/todo.php
+++ b/script/todo.php
@@ -27,7 +27,6 @@ class Storage
 
     public function store(string $path, string $content): void
     {
-        $path = $this->realpath($path);
         file_put_contents($path, $content);
     }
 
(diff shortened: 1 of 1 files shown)


script/todo.php
```diff
@@ -27,6 +27,7 @@ public function load(string $path)
 
     public function store(string $path, string $content): void
     {
+        $path = $this->realpath($path);
```
andrey-helldar, line 30: Remove this line.

Its problem is that it calls the realpath method. If we delete the todo.md file, the method will return `null` and an error will occur, because the path parameter of the `file_put_contents` function cannot be null.

See https://github.com/Laravel-Lang/lang/pull/1401/commits/deb228567cec25e33eb99d8db64a2557b3abf437

andrey-helldar (changes requested):

script/todo.php
olsza: thx

andrey-helldar (approved):

caouecs:
Thank you

PR #1407, opened 2020-10-08
Update nl.json

Lubje:
Remove initial English variant for "Forgot Your Password?".

### DIFF
--- a/json/nl.json
+++ b/json/nl.json
@@ -13,7 +13,6 @@
     "Error": "Fout",
     "Forbidden": "Geen toegang",
     "Forgot your password? No problem. Just let us know your email address and we will email you a password reset link that will allow you to choose a new one.": "Forgot your password? No problem. Just let us know your email address and we will email you a password reset link that will allow you to choose a new one.",
-    "Forgot your password?": "Forgot your password?",
     "Forgot Your Password?": "Wachtwoord vergeten?",
     "Go Home": "Terug naar de voorpagina",
     "Hello!": "Hallo!",


luisprmat:
@Lubje You should **NOT delete** this key because it is a current [translation](https://github.com/laravel/jetstream/blob/1.x/stubs/resources/views/auth/login.blade.php#L38) in laravel Jetstream.

Remember that the JSON keys are *case sensitive* therefore `'Forgot your password?'` is different to `'Forgot Your Password?'`.

Therefore I consider that neither of the two keys should be deleted or modified because both are used:
* `'Forgot your password?'` in [Laravel Jetstream](https://github.com/laravel/jetstream/blob/1.x/stubs/resources/views/auth/login.blade.php#L38) and 
* `'Forgot Your Password?'` in [Laravel UI](https://github.com/laravel/ui/blob/3.x/src/Auth/bootstrap-stubs/auth/passwords/confirm.stub#L38)

If we delete the second we would lose the translation for those who still use Laravel UI

@caouecs I hope I have explained the reasons why this PR must be fixed before merging

json/nl.json
```diff
@@ -13,7 +13,6 @@
     "Error": "Fout",
     "Forbidden": "Geen toegang",
     "Forgot your password? No problem. Just let us know your email address and we will email you a password reset link that will allow you to choose a new one.": "Forgot your password? No problem. Just let us know your email address and we will email you a password reset link that will allow you to choose a new one.",
-    "Forgot your password?": "Forgot your password?",
```
luisprmat, line 16: @Lubje You should **NOT delete** this key because it is a current [translation](https://github.com/laravel/jetstream/blob/1.x/stubs/resources/views/auth/login.blade.php#L38) in laravel Jetstream.

Remember that the JSON keys are *case sensitive* therefore `'Forgot your password?'` is different to `'Forgot Your Password?'`.

Therefore I consider that neither of the two keys should be deleted or modified because both are used:
* `'Forgot your password?'` in [Laravel Jetstream](https://github.com/laravel/jetstream/blob/1.x/stubs/resources/views/auth/login.blade.php#L38) and 
* `'Forgot Your Password?'` in [Laravel UI](https://github.com/laravel/ui/blob/3.x/src/Auth/bootstrap-stubs/auth/passwords/confirm.stub#L38)

If we delete the second we would lose the translation for those who still use Laravel UI

I suggest translate this key instead of delete

luisprmat (changes requested):

Lubje:
I understand, thanks for the explanation, appreciate it!

caouecs:
@Lubje you can help this project with other need translations https://github.com/Laravel-Lang/lang/blob/master/todo.md#nl ;)

@luisprmat 👍

Lubje:
I'll  look into adding these missing translations tomorrow for the nl.json file.

PR #1409, opened 2020-10-08
Update de.json

ChristofK:
Changed "Remember Me" to "Remember me"

### DIFF
--- a/json/de.json
+++ b/json/de.json
@@ -37,7 +37,7 @@
     "Please confirm your password before continuing.": "Bitte bestätigen Sie Ihr Passwort bevor Sie fortfahren.",
     "Regards": "Mit freundlichen Grüßen",
     "Register": "Registrieren",
-    "Remember Me": "Angemeldet bleiben",
+    "Remember me": "Angemeldet bleiben",
     "Resend Verification Email": "Bestätigungslink erneut senden",
     "Reset Password Notification": "Benachrichtigung zum Zurücksetzen des Passworts",
     "Reset Password": "Passwort zurücksetzen",


luisprmat:
@ChristofK  You should **NOT change** this key because it is a current [translation in Laravel UI](https://github.com/laravel/ui/blob/3.x/src/Auth/bootstrap-stubs/auth/login.stub#L48).

Remember that the JSON keys are *case sensitive* therefore `'Remember me'` is different to `'Remember Me'`.

Therefore I consider that neither of the two keys should be deleted or modified because both are used:
* `'Remember me'` in [Laravel Jetstream](https://github.com/laravel/jetstream/blob/1.x/stubs/resources/views/auth/login.blade.php#L31) and 
* `'Remember Me'` in [Laravel UI](https://github.com/laravel/ui/blob/3.x/src/Auth/bootstrap-stubs/auth/login.stub#L48)

If you modify that key as you did we would lose the translation for those who still use Laravel UI

I suggest you keep them both and translate them if necessary with the same text.

json/de.json
```diff
@@ -37,7 +37,7 @@
     "Please confirm your password before continuing.": "Bitte bestätigen Sie Ihr Passwort bevor Sie fortfahren.",
     "Regards": "Mit freundlichen Grüßen",
     "Register": "Registrieren",
-    "Remember Me": "Angemeldet bleiben",
+    "Remember me": "Angemeldet bleiben",
```
luisprmat, line 40: I suggest you keep both keys so it seems repeated

luisprmat (changes requested):

luisprmat:
@ChristofK I invite you to collaborate in this project by adding and translating the missing keys in [Todo List](https://github.com/Laravel-Lang/lang/blob/master/todo.md#de)

andrey-helldar:
@ChristofK, see also https://github.com/Laravel-Lang/lang/pull/1335

json/de.json
ChristofK: Yes, you are right.

PR #1410, opened 2020-10-09
Feature - Translate some Laravel 8 fields

alancolant:
ajout traductions francaises


alancolant:
Change already present in latest branch

PR #1417, opened 2020-10-17
[es] Complete translations

luisprmat:
Added missing translation for `validation-inline` and added an exclusion


caouecs:
Gracias

PR #1422, opened 2020-10-19
[en, es] Update wording translation

luisprmat:
Taylor has updated this translation
[profile/logout-other-browser-sessions-form.blade.php](https://github.com/laravel/jetstream/commit/055be8529b4fb45897e9d5ef42eee0b2a45c8bde)

### DIFF
--- a/json/es.json
+++ b/json/es.json
@@ -58,7 +58,7 @@
     "Go to page :page": "Ir a la página :page",
     "Hello!": "¡Hola!",
     "hi": "hey",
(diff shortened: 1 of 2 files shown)


caouecs:
We consider this tag as an updated tag or a new tag ?

luisprmat:
I think we should modify it because if a user already has an existing project and updates jetstream their translations will not be affected if he does not overwrite the views. Also if someone installs jetstream they usually do it on a new project. So I don't think these two tags should coexist

caouecs:
Ok, I merge your pull request and after I update all language with this tag

caouecs:
Gracias

PR #1429, opened 2020-10-24
add some attributes to fa lang

aminireza-ir:
I tried to add some attributes that used too many in our project


andrey-helldar:
It is not at all necessary to add third-party keys to the shared project.
You can add translation keys to your validation.php file.
Use a translation manager such as [andrey-helldar/laravel-lang-publisher](https://github.com/andrey-helldar/laravel-lang-publisher) to update other keys. It saves your keys while updating files. Thus, you do not need to worry about data safety.

PR #1442, opened 2020-11-13
[en, es] Add new tag in auth.php

luisprmat:
https://github.com/laravel/laravel/blob/8.x/resources/lang/en/auth.php#L17

New tag added to `auth.php`


caouecs:
Gracias

PR #1444, opened 2020-11-15
[ko] Translate json with some validation

SilNex:


src/ko/validation-inline.php
```diff
@@ -90,7 +90,7 @@
         'string'  => '이 문자열의 길이는 최소한 :min자이어야 합니다.',
         'array'   => '이 배열의 항목 수는 최소한 :min개의 항목이 있어야 합니다.',
     ],
-    'multiple_of'          => 'The value must be a multiple of :value',
+    'multiple_of'          => ':attribute 는 :value 의 배수여야 합니다.',
```
caouecs, line 93: No :attribute here

caouecs (changes requested):
Thank you, just a typo

SilNex:
fixit !

caouecs:
Thank you

PR #1450, opened 2020-11-19
Updated passwords.php

unknownbreakdown:
Translations have been slightly modified in order to follow the same formal pattern of translations in other files.

### DIFF
--- a/src/es/passwords.php
+++ b/src/es/passwords.php
@@ -12,9 +12,9 @@ return [
     |
     */
 
-    'reset'     => '¡Tu contraseña ha sido restablecida!',
-    'sent'      => '¡Te hemos enviado por correo el enlace para restablecer tu contraseña!',
-    'throttled' => 'Por favor espera antes de intentar de nuevo.',
-    'token'     => 'El token de recuperación de contraseña es inválido.',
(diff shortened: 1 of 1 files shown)


luisprmat (approved):
These changes are consistent, but in this issue, it is good to clarify that in Spanish there are two ways of expressing oneself in relation to the pronoun ***you*** (second person).
The first one uses **usted** and is [more formal](https://www.maidertomasena.com/como-decidir-si-escribir-de-tu-o-de-usted-en-la-web-de-tu-negocio/), the second uses the pronoun ***tú*** and it is [more friendly](https://es.wikipedia.org/wiki/Tuteo) and generates more confidence and is also quite used.
In this project we are using the first way in most translations (with these corrections I hope in all).
But I agree to consider the possibility of opening a second *set* of translations for Spanish where we only use the second way ***tú***, @caouecs I don't know if this is possible and what would be the best way to do it.

caouecs:
Of course, it's possible to have es-usted and es-tu versions

but it's a double work ;)

luisprmat:
There is also no problem in merging this PR

caouecs:
Gracias

PR #1457, opened 2020-11-26
Added all missing translation for nl

Loots-it:
I added all missing translations for nl. I didn't change the todo.md but I think the script runs automatically?


caouecs:
Thank you, and yes for todo

PR #1459, opened 2020-11-27
Updated italian translations

masterix21:
Added:
- password in auth.php
- "This is a secure area of the application. Please confirm your password before continuing." in it.json


caouecs:
Grazie

PR #1463, opened 2020-11-30
[sr_Latn_ME] Translation fix.

milose:
Translated and fixed most of Montenegrin language strings.


src/sr_Latn_ME/pagination.php
```diff
@@ -12,6 +12,6 @@
     |
     */
 
-    'previous' => '&laquo; Prethodna',
-    'next'     => 'Sledeća &raquo;',
+    'previous' => '&laquo; prethodna',
+    'next' => 'sljedeća &raquo;',
```
caouecs, line 16: Don't change indentation

caouecs (changes requested):
thank you, just a problem on indentation

src/sr_Latn_ME/pagination.php
milose: Sorry about that.

caouecs:
Thank you

PR #1467, opened 2020-12-07
[fa] Update auth, validation and validation-inline

meysamzandy:
Update auth.php for password in auth
Update validation-inline.php for multiple_of in validation-inline
Update validation.php for multiple_of in validation
Update todo.md for fa


caouecs:
Thank you

PR #1479, opened 2021-01-03
[sl] Added missing translations everywhere

harrallo:
Not able to update todo. But there should not be any missing translations in sl (Slovenian) any more. At least up until now.


caouecs:
Thank you

caouecs:
Hello,
can you add "Administrator" and "Email" on json file ? If it's the same translation, you can add a exclude file ( see [Contributing doc](https://github.com/Laravel-Lang/lang/blob/master/CONTRIBUTING.md#what-should-i-do-if-there-is-a-tag-whose-translation-is-the-same-as-in-english) )

harrallo:
Hi,

I did. Both could be on the exclusion list, I just forgot about them since
there were so many translations. Anyhow I used more formal translation of
Email and added Administrator to exclusion list.
Thank you for the email. And for updating todo.

Wanted to fix translations before starting a new Laravel 'for hobby'
project after a long time. Have been in the .NET world for the last few
years. With SVN. Yuck.

Natalia

PR #1480, opened 2021-01-04
[sl] Updated 2 missing translations

harrallo:
'Email' changed to 'Elektronski naslov' (both are ok in sl)
Added 'Administrator' to exclusion list

'todo' file has to be updated by somebody else. Thank you.


caouecs:
The todo file is generated by Github itself ;)

caouecs:
Thank you

PR #1488, opened 2021-01-09
[en, es] Add missing tags related to button's labels for mail - team …

luisprmat:
Button's labels in the mail template for team invitations are pending to translate

- [Create Account](https://github.com/laravel/jetstream/blob/2.x/resources/views/mail/team-invitation.blade.php#L7)
- [Accept Invitation](https://github.com/laravel/jetstream/blob/2.x/resources/views/mail/team-invitation.blade.php#L13)

### DIFF
--- a/src/es/validation.php
+++ b/src/es/validation.php
@@ -174,6 +174,7 @@ return [
         'password_confirmation' => 'confirmación de la contraseña',
         'phone'                 => 'teléfono',
         'price'                 => 'precio',
+        'role'                  => 'rol',
         'second'                => 'segundo',
         'sex'                   => 'sexo',
         'subject'               => 'asunto',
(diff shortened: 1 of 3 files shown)


caouecs:
Gracias

mostafa2121:
[image: Screenshot from 2021-01-07 17-45-08.png]

src/es/validation.php
```diff
@@ -174,6 +174,7 @@
         'password_confirmation' => 'confirmación de la contraseña',
         'phone'                 => 'teléfono',
         'price'                 => 'precio',
+        'role'                  => 'rol',
```
mohamedsabil83, line 177: **ICYMI**: **role** was mentioned in json file.

luisprmat: I added this attribute (`role`) to the `validation.php`  file so that I can correctly translate the error message when I don't select any role when adding a new member to the team. Otherwise you would get the **El campo role es obligatorio.** message which is incorrect in Spanish. The correct message is **El campo rol es obligatorio.**

RoleValidation

PR #1492, opened 2021-01-11
[sv] Cleaning and adding

hantrick:
Removed unusued strings and fixed incorrect strings. Also added the basic strings that's needed for most Laravel apps.


caouecs:
Thank you

PR #1496, opened 2021-01-12
Update validation.php with 100 percent accuracy.

g-e-g-e-r-s-o-n:


src/ka/validation.php
```diff
@@ -12,111 +13,111 @@
     |
     */
 
-    'accepted'             => ':attribute უნდა იყოს მონიშნული.',
```
caouecs, line 15: Don't update space. Thanks

caouecs (changes requested):

PR #1500, opened 2021-01-12
Bosnian translation

amiranagram:
I added the missing Bosnian translation and fixed some others to be more meaningful. I'll translate the JSON file when I have more time.


caouecs:
Thank you

In auth.php, can you add "'password' => 'The provided password is incorrect.'," ?

amiranagram:
Yes I will. 

I was meaning to ask what's the purpose of validation-inline.php file inside src folders?

caouecs:
validation-inline.php is the same translations than validation.php but without the name of field

PR #1505, opened 2021-01-13
Bosnian translation (add missing password translation in auth.php)

amiranagram:
Missing translation, as promised.


caouecs:
Thank you

PR #1506, opened 2021-01-13
[en, es] Translate cancel button of team invitation

luisprmat:
This modification was ready as of January 6 and had been ignored
- [Cancel](https://github.com/laravel/jetstream/blob/2.x/stubs/livewire/resources/views/teams/team-member-manager.blade.php#L102)


caouecs:
Gracias

PR #1515, opened 2021-01-20
[ru] Add translate for `current_password`

russsiq:
Add translate for `validation.attributes.current_password`.

This attribute is used in the **Update Password** form of the `/user/profile` route from the `laravel/fortify` package.


andrey-helldar (approved):
Yes, there is a code like this: https://github.com/laravel/fortify/blob/1.x/stubs/UpdateUserPassword.php#L23

caouecs:
Спасибо

PR #1519, opened 2021-01-23
Update ka.json

CSenshi:
This pr copies translations from src/ka to ka.json. 

They are same lang translations so why not have translations that are in .php to .json format as well? Am I missing something?


andrey-helldar:
You don't need to do this.

Translations of validation are in the `validation.php` and `validation-inline.php` files.

Why do you want to duplicate data?

andrey-helldar (commented):
You don't need to do this.

Translations of validation are in the `validation.php` and `validation-inline.php` files.

Why do you want to duplicate data?

You don't use `__('The :attribute field must be true or false.')` in your code, do you? Right?

PR #1526, opened 2021-02-09
[de] Correct some grammatical errors and add a second unformally translation

michaelgrossklos:
[de] Correct some grammatical errors and add a second unformally translation. Because, in German there is a formal and a unformal salutaiton. They get differentiated by `de_DE_sie` (formal) and `de_DE_du` (unformal).


andrey-helldar:
Provided localization codes are based on [ISO-639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and [ISO-15897](https://laravel.com/docs/8.x/localization). I think it is not worth making different variations of localizations that are not related to these codes.

In this regard, it would be appropriate to keep the formal version of the messages.

caouecs:
I agree with andrey, languages have many variations according to the people you speak but in this package, the best, for me, it's a classic version, each one can then customize then

andrey-helldar:
I think that if you need such differences in translation, not related to official languages, then you can store them in your package and keep them up to date. And if this translation is needed in several of your projects, then you can create your own package for this. For example using the [package wizard](https://github.com/andrey-helldar/package-wizard).

PR #1529, opened 2021-02-16
[nl] Added missing translations and set language style to informal

jaapboerma:
I have added the missing translations for the Dutch language. In the existing file, a mix of formal and informal language was used. I chose to align the entire translation to informal, since that feels more appropriate. I also ran the `scripts/todo.php` in order to make sure all translations exist. This results in the `todo.md` being added to this PR. Hope that's not a problem, feel free to remove it.


caouecs:
Dankjewel

PR #1534, opened 2021-02-23
[en, es] Replace words 'Login' and 'Logout' by 'Log in' and 'Log Out'

luisprmat:
Recently there were changes in Laravel Jetstream that may affect translations. [The words 'Login' and 'Logout' were separated into 'Log in' and 'Log out'](https://github.com/laravel/jetstream/commit/e85dc61adb78c2c63a9ac6e174adaaafa9003d8d). I make this update keeping the above to preserve backwards compatibility, but I think the previous translations should be removed.


luisprmat:
[Laravel Breeze](https://github.com/laravel/breeze/blob/1.x/stubs/default/resources/views/layouts/navigation.blade.php#L44) uses *Log out* instead *Log Out*.

caouecs:
grazias

PR #1540, opened 2021-02-28
[en, ru] Jetstream: Logout > Log Out

andrey-helldar:


caouecs:
I just have a problem with the retro-compatibility :)

but if everybody is ok...

andrey-helldar:
Need to think :)

andrey-helldar:
@caouecs, I prefer to keep these keys as backward compatibility until we publish version 9.0 😊

PR #1544, opened 2021-03-09
[zh_*] Translate missing lines.

fnoopv:
- add missing 7 translation
- fix `Log Out Other Browser Sessions` translation
- adjust `If necessary, you may logout of all of your other browser sessions across all of your devices. Some of your recent sessions are listed below; however, this list may not be exhaustive. If you feel your account has been compromised, you should also update your password.`  translation


caouecs:
Thank you

andrey-helldar:
@fnoopv, we have added a new key for Fortify (#1545). Can I ask you to translate it too? :)

`The :attribute must be at least :length characters and contain at least one special character and one number.`

image

fnoopv:
ok, i will translate it

PR #1545, opened 2021-03-09
[en, es] Add missing scenario in Password Rule (Fortify)

luisprmat:
From [Jan 28](https://github.com/laravel/fortify/commit/3ddabcaf4e339a85b86f2ff45dac7905ff8840d8) is missing an new scenario in `Rules\Password.php`:
- [The :attribute must be at least :length characters and contain at least one special character and one number.](https://github.com/laravel/fortify/blob/1.x/src/Rules/Password.php#L128)


caouecs:
Gracias

PR #1546, opened 2021-03-09
[ru] Add missing key in Password Rule (Fortify)

andrey-helldar:
Also see https://github.com/Laravel-Lang/lang/pull/1545


caouecs:
Спасибо

PR #1550, opened 2021-03-11
[en, es] Add translations for Laravel Cashier

luisprmat:
These translations are included in the following files:
- [`resources/views/payment.blade.php`](https://github.com/laravel/cashier-stripe/blob/12.x/resources/views/payment.blade.php)
- [`src/Notifications/ConfirmPayment.php`](https://github.com/laravel/cashier-stripe/blob/12.x/src/Notifications/ConfirmPayment.php)

In [Laravel Cashier](https://laravel.com/docs/8.x/billing)


caouecs:
Gracias

PR #1554, opened 2021-03-12
[en, ru] Added Laravel Nova key translations

andrey-helldar:
PS: I chose not to translate the country names, so I created a global exception file and added the countries to it.

Also see https://github.com/Laravel-Lang/lang/discussions/1547


andrey-helldar:
@luisprmat, this is both yes and no.
Yes, because a lot of keys are added.
No, because they do not provide backward compatibility, they only expand the scope.

caouecs:
thank you... we have a lot of work now...

andrey-helldar:
Yep)

PR #1559, opened 2021-03-15
[zh_*] add missing translation

fnoopv:
- add missing 16 translation
- fix `The :attribute must be a valid role.` translation


caouecs:
Thank you

PR #1560, opened 2021-03-15
Added check for the existence of keys

andrey-helldar:
What I've done:

1. Added a script for checking the existence of keys. For example, if we remove the "foo" key from an English file, then it will be removed from all localizations. If you add the "bar" key to an English localization, it will be added to all localizations;
2. Updated script for GitHub Actions;
3. In dev-dependencies, added packages for easy work with files (most of the functionality is implemented in them). These packages WILL NOT be installed by users. Local development only. In our case, in GitHub Actions;
4. Removed the search for unknown keys (`: including not present`) from the todo.md file generator;
5. In this PR I included changes only for French localization, so as not to create noise;
6. Sorting keys under the hood is done;
7. Added a code that sorts the keys of the English localization (without this, the tests will not work correctly, and it will be easier for us);
8. Added unit tests for checking keys for all pushes (Tests for PHP are specially divided by file names so that in case of errors we know which file the problem is with ([see](https://github.com/Laravel-Lang/lang/actions/runs/654706550)));

Now the tests are being executed with an error for the reason that in this PR I did not commit the fixes for other languages.

image

image

@Laravel-Lang/laravel-lang, guys.


src/fr/auth.php
```diff
@@ -1,17 +1,6 @@
 <?php
 
 return [
-    /*
```
caouecs, line 4: You delete comments ?

src/fr/validation-inline.php
```diff
@@ -115,67 +103,12 @@
...
-    | Custom Validation Attributes
-    |--------------------------------------------------------------------------
-    |
-    | The following language lines are used to swap attribute place-holders
-    | with something more reader friendly such as E-Mail Address instead
-    | of "email". This simply helps us make messages a little cleaner.
-    |
-    */
-
-    'attributes' => [
```
caouecs, line 150: You delete attributes ?

script/en/validation-inline.php
```diff
@@ -31,6 +19,11 @@
     ],
     'boolean'              => 'This field must be true or false.',
     'confirmed'            => 'The confirmation does not match.',
+    'custom'               => [
```
caouecs, line 22: Custom here, it's not good

script/en/validation.php
```diff
@@ -57,8 +52,8 @@
     ],
     'image'                => 'The :attribute must be an image.',
     'in'                   => 'The selected :attribute is invalid.',
-    'in_array'             => 'The :attribute field does not exist in :other.',
     'integer'              => 'The :attribute must be an integer.',
+    'in_array'             => 'The :attribute field does not exist in :other.',
```
caouecs, line 56: Your script sorts 'in_array' after 'integer'

caouecs (changes requested):
Your dev is interesting but I have some questions about sort and delete of comments

src/fr/auth.php
andrey-helldar: I [said in the discussion](https://github.com/Laravel-Lang/lang/discussions/1555) that in this case all comments will be automatically deleted. The code works with an array, and does not parse strings.

src/fr/validation-inline.php
andrey-helldar: No, it has been moved alphabetically.

script/en/validation-inline.php
andrey-helldar: It has been moved alphabetically.

script/en/validation.php
andrey-helldar: `ksort($array)`:
```
API key
all key
foo_key
fookey
```

`ksort($array, SORT_FLAG_CASE ^ SORT_NATURAL)`:
```
all key
API key
fookey
foo_key
```

src/fr/validation-inline.php
andrey-helldar: I figured out how to apply the correct sorting. I will do it in the coming days.

script/en/validation-inline.php
andrey-helldar: I figured out how to keep the location of the `custom` and` attributes` keys at the end of the file. I will do it in the near future.

script/en/validation-inline.php
```diff
@@ -58,8 +51,8 @@
     ],
     'image'                => 'This must be an image.',
     'in'                   => 'The selected value is invalid.',
-    'in_array'             => 'This value does not exist in :other.',
     'integer'              => 'This must be an integer.',
+    'in_array'             => 'This value does not exist in :other.',
```
mohamedsabil83, line 55: What kind of sort will we adopt for **underscore**?
become first or last?

script/en/validation.php
mohamedsabil83: I see. I review that above but find the reason here.

andrey-helldar:
I will clarify some points:

1. Comments are removed from the final files because the code reads them as an array, not as strings, and works like an array. The array, unfortunately, cannot contain comments;
2. Keys are now sorted by passing flags ([SORT_FLAG_CASE ^ SORT_NATURAL](https://www.php.net/manual/en/function.ksort)), but I can specify my callback for the sorting rules. You just need to decide which one.

Here are some standard options:

image
image
image

andrey-helldar:
image

Also added sorting of nested arrays:

image

src/fr/auth.php
andrey-helldar: For a long time, the Russian-speaking community has not paid attention to the comments inside the translation files, since the name of the file already makes it clear what messages are inside.

script/en/validation.php
andrey-helldar: With the latest changes, the sort looks like this:

image

andrey-helldar:
Found and fixed a bug with sorting:

image

mohamedsabil83:
First, thank you for your great work.

I vote for the last choice and will be better if can sort **underscore** after **hyphen**.
> image

caouecs:
I will review the pull request this week-end, problem at work and with OVH :)

andrey-helldar:
@caouecs, I will remove the WIP status from the title when the PR is ready and it suits all of us :)

andrey-helldar:
@mohamedsabil83, @caouecs, see, will this sorting option suit you?

image

Personally, I like it :)

PS: In our case, numeric keys are not used, so you can ignore them.

mohamedsabil83:
Yes, I like it too, even with sorting the numeric at the end. 👍🏽

andrey-helldar:
There are no numeric keys in our project, so you can ignore them.
In the example, I added that there was an objective assessment. 😊

mohamedsabil83:
Yes, I mean that I support this sorting even if there are numeric keys added in the future.

andrey-helldar:
I did the sorting in the following order: special characters, numbers, strings:

image

😊

If someone wants to add to their application, then the method lies here: [Helldar\Support\Facades\Helpers\Arr::ksort()](https://github.com/andrey-helldar/support/blob/main/src/Helpers/Arr.php#L143-L164).

andrey-helldar:
@caouecs, ready to release 😊

andrey-helldar:
PR: tests fail with an error because in my PR the keys of all localizations, except for the French one, are not sorted. Let me remind you that this was done on purpose so as not to create noise in PR.

I also changed the GitHub Actions - when pushing to the "master" branch, keys will be automatically sorted and pushed to the repository.

andrey-helldar:
@caouecs, sorry, not ready yet.
Now I checked how GitHub Actions work and found this bug:

`$ php script/keys.php`

image

`$ php script/keys.php`

image

andrey-helldar:
@caouecs, sorry if you tired of my messages :)

Removed the quote character from the file - it was the bug trigger.

And I will correct the reason for this processing in the near future, it is processing in my package.

image

mohamedsabil83 (approved):
Just reviewed. It sorts keys as expected. Also, all test units passed successfully.

andrey-helldar:
@caouecs, hi! Did you manage to see the code?

caouecs:
it's in my todo list :innocent:

I have a problem with your test on PHPUnit, a fail ;)

andrey-helldar:
> it's in my todo list 😇
> 
> I have a problem with your test on PHPUnit, a fail ;)

If you just take my PR and run the test, there will be a 100% error. This is because before the test, you must execute `php script/keys.php` and `php script/todo.php`.

I deliberately did not include in this PR changes to files of other languages, so as not to create noise.

image

When the PR is accepted, GitHub Actions will automatically run these scripts ([see](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-15-15-25/.github/workflows/make-todo.yml#L40-L62)).

caouecs:
let's go

PR #1562, opened 2021-03-21
[de] catch up with jetstream update

j2L4e:
Unbreaks some translations for phrases changed via laravel/jetstream#677


andrey-helldar:
We do not remove old keys like "login" and "logout" to maintain backward compatibility, but we really appreciate your efforts to make this world a better place! Thank you!

PR #1565, opened 2021-03-24
[tr] Add prohibited validation rule

mertasan:
laravel/framework#36667
laravel/docs#6931
laravel/laravel#5569


andrey-helldar (changes requested):
Also add these keys to the [script/en/validation.php](https://github.com/Laravel-Lang/lang/blob/master/script/en/validation.php) and [script/en/validation-inline.php](https://github.com/Laravel-Lang/lang/blob/master/script/en/validation-inline.php) files.

If this is not done, then upon acceptance of the PR, the script will automatically delete the keys that do not correspond to the validation.php file.

caouecs:
Thank you

PR #1566, opened 2021-03-25
fix missing translations at es.json

fcolecumberri:


json/es.json
```diff
@@ -1,71 +1,71 @@
...
     "Aruba": "Aruba",
-    "Attach": "Attach",
-    "Attach & Attach Another": "Attach & Attach Another",
-    "Attach :resource": "Attach :resource",
-    "August": "August",
+    "Attach": "Vincular",
+    "Attach & Attach Another": "Adjuntar & adjuntar otro",
+    "Attach :resource": "Adjuntar :resource",
+    "August": "Agosto",
     "Australia": "Australia",
```
caouecs, line 69: Without change, can you add these tags in script/excludes/es.php ?

caouecs (changes requested):
Hola, 
Grasias.

But we need update the exclude files, if there is no change between tag and translation

json/es.json
andrey-helldar: @caouecs, it's not obligatory. I brought the names of the countries into the global file of exceptions: [script/excludes/_all.php](https://github.com/Laravel-Lang/lang/blob/master/script/excludes/_all.php).

json/es.json
```diff
@@ -1,71 +1,71 @@
...
-    "Add Team Member": "Agregar nuevo miembro al equipo",
-    "Added.": "Agregado.",
+    "Add row": "Añadir fila",
+    "Add Team Member": "Añadir un nuevo miembro al equipo",
+    "Added.": "Añadido.",
     "Administrator": "Administrador",
     "Administrator users can perform any action.": "Los administradores pueden realizar cualquier acción.",
     "Afghanistan": "Afghanistan",
-    "Aland Islands": "Åland Islands",
+    "Aland Islands": "Islas Aland",
```
andrey-helldar, line 27: Country names do not have to be translated into your language. This can only be done if you need names in your language.

caouecs: I prefer to translate countries. I prefer "Espagne" than "Spain", for example

andrey-helldar: I understood. Look, when adding new keys, they are automatically added to all localizations, but since they are specified in the file for ignore, their translation is not required.

For example, in Russian localization, it is not at all necessary to translate country names into Russian. We often use English versions.

fcolecumberri: If someone doesn't want to translate the name of a country they just can put the name outside the `__()` I mean if you write:

```blade
{{__('Country')}}: {{__('Mexico')}}
```
Into Spanish it should render (Notice the **é** is not **e**):
```
País: México
```

If you don't want to translate it, you should do:

```blade
{{__('Country')}}: Mexico
```

So it would render:
```
País: Mexico
```

There is **literally no point** making a translation file that does not translate

andrey-helldar: I disagree. It is a bad idea to prescribe names in the code, because someone needs a translation, and someone does not.
For example, in Russian-language localization, the list of countries is rarely translated into Russian. Therefore, "Mexico" is displayed as "Mexico".

json/es.json
fcolecumberri: I don't understand what you want me to do, I understand the English, but I am not familiarize with the project and I maybe be miss reading the logic of the project.

What is the `script/excludes/es.php` suppose to do?

(please don't tell me it excludes the translation from happening, that would defeat the porpoise of having the entry in the json file in the first place)

json/es.json
fcolecumberri: If the Russian language has it's grammar exceptions that makes it untranslatable, that should be only on the Russian json file. not because English→Russian is untranslatable it means English→Spanish is not.

andrey-helldar: It seems to me that we did not understand each other.
I am saying that translation is not required if you do not need it.

fcolecumberri: actually if Russian can't translate, the ´ru.json´ shouldn't have the country key on the first place.

json/es.json
andrey-helldar: You don't have to do anything.

This file does not exclude translation. It is needed to correctly form the todo.md file.

json/es.json
fcolecumberri: In Spanish you need to translate the country names. As someone that has lived on Mexico and Spain for years I ensure you must Spanish speakers only know those country names in Spanish.

andrey-helldar: Yes and no.

One of the old problems with this project is that the translation files contain outdated keys, which causes the problem of their validation. We have recently introduced the functionality of automatic key checking in all localizations. This code removes unused keys and adds missing ones.

json/es.json
fcolecumberri: ok

json/es.json
fcolecumberri: If someone is writing a code and write 
```blade
{{__('Country')}}: {{__('Spain')}}
```
Give me a reason for the **Spanish** render (A.K.A. not Russian) it should render:
```
País: Spain
```
instead of
```
País: España
```
In Spanish translating the country name makes a lot of sense.

andrey-helldar: Are we discussing Taylor's solution to add code.

I repeat once again: if you want to translate, translate. Nobody forbids.

I see no reason to discuss this topic.

andrey-helldar (approved):

PR #1568, opened 2021-03-25
Update it.json

masterix21:


json/it.json
```diff
@@ -247,60 +247,60 @@
...
     "Jersey": "Jersey",
-    "Jordan": "Jordan",
-    "July": "July",
-    "June": "June",
-    "Kazakhstan": "Kazakhstan",
+    "Jordan": "Giordania",
+    "July": "Luglio",
+    "June": "Giugno",
+    "Kazakhstan": "Kazakistan",
     "Kenya": "Kenya",
```
caouecs, line 268: In exclude file ?

caouecs (changes requested):
Hello,
just a question on exclude tag

json/it.json
masterix21: Many countries have no translations. English "Kenya" is the same in Italian.

masterix21: If you like, I can add these exclusions to `script/excludes/it.php`.

caouecs: yes ;)

masterix21: Done ;)

script/excludes/it.php
```diff
@@ -11,9 +11,135 @@
...
+    'ID',
     'Email',
     'Oh no',
     'Editor',
     'Password',
     'Dashboard',
+    'No',
+
+    /**
+     * Countries
```
andrey-helldar, line 23: The country names do not need to be specified in the exclusion list, which is needed to generate the todo.md file.

The list of countries is already listed in the global list for all localizations: [script/excludes/_all.php](https://github.com/Laravel-Lang/lang/blob/master/script/excludes/_all.php)

andrey-helldar (changes requested):

script/excludes/it.php
caouecs: ok

caouecs:
Grazie

PR #1577, opened 2021-03-31
Improving generation scripts (todo and keys)

andrey-helldar:
1. Optimized the location of the scripts:
    * `php app/keys.php`
    * `php app/todo.md`
2. The main TODO file is in the same location: `TODO.md`;
3. TODO localization files are in the `todo` folder;
4. The executable code was made according to PSR-4;
5. When installing a package into an application, script files will not be exported (see `.gitattributes`);
6. Moved the `script/excludes` folder to the root of the project;
7. PHP 8 is required to run `app/todo.php` and `app/keys.php`

**Show [live view](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-26/22-16/TODO.md).**

image

image

image


caouecs:
Why PHP 8 ?

andrey-helldar:
1. Strong multiple typing was needed in some places. For example, [here](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-26/22-16/app/main/Contracts/Filesystem.php#L13).
2. PHP 8 made it possible to shorten a part of the code. For example, [here](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-26/22-16/app/main/Models/Locale.php#L7-L11).
3. The class attributes have explicit typing. This allows not only to control its content, but also there is no need to specify dock blocks for the IDE. For example, [here](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-26/22-16/app/main/Processors/Processor.php#L15-L19).
4. This code is only needed to check the keys and generate the TODO.md file - the PHP version has no effect on developer applications (the version is indicated in the [require-dev](https://github.com/andrey-helldar/Laravel-lang/blob/patch/2021-03-26/22-16/composer.json#L24-L25) block).
5. GitHub Actions can work with PHP 8, so there will be no problems with it.
6. PHP 8 was released half a year ago and why not use it?

caouecs:
I just ask because I did not see specific code for PHP8, but now it's more clear ;)

andrey-helldar:
> I just ask because I did not see specific code for PHP8, but now it's more clear ;)

Specific code for PHP8 here:

image

image

image

image

image

image

image

image

image

andrey-helldar:
Installed without problems 🙂

image

caouecs:
I'm ok for multi types, but for simple type, I'm sorry, I work on PHP 7.4, and I can

PR #1578, opened 2021-03-31
Updated tests

andrey-helldar:
Just PHP 8 style tests 😊


caouecs:
in PHP 7, you can define the type of variable

thank you

PR #1579, opened 2021-04-01
[es] Add missing translations and full ortography review.

luisprmat:
Several translations for *Laravel Nova* were wrong! There may be more but we will fix them when they are detected.
Good job! Thanks specially to @andrey-helldar for refactoring of this package.


caouecs:
Gracias

PR #1583, opened 2021-04-01
The list of countries has been removed from ignoring when generating todo.md

andrey-helldar:


luisprmat (commented):
I guess this means that I still have to add some countries to my exclusion list?

andrey-helldar:
> I guess this means that I still have to add some countries to my exclusion list?

I think yes. The list of countries is now on the global list of exclusions. Those names of countries that in your language coincide with the English version must be specified in the file `excludes/<lang>.php`.

caouecs:
Спасибо

PR #1600, opened 2021-04-08
Added validation for the `:attribute` string in the `validation-inline.php` files

andrey-helldar:
image

image


caouecs:
thank you

PR #1604, opened 2021-04-10
Added check of exclusion files for keys relevance

andrey-helldar:
* Added check of exclusion files for keys relevance
* Added templates for saving files

Before > After:

> Excludes file
>
> image

> Locale's file
>
> image

What problems have been solved:

1. Duplicate keys were specified in some exclusion files;
2. There was no sorting in the exclusion files;
3. Some keys were excluded from the main files.

What's happening:
1. Values are sorted alphabetically;
2. Keys deleted from EN files are excluded (both php and json files are checked);
3. A comment is inserted at the beginning of each file.

Also, when processing translation files, a comment is now inserted with the description of this file (`app/resources/arrays/*.stub`).

The commentary is indicated in English only and is for information only. It is NOT NECESSARY to change the comment text directly in the file - it will be automatically replaced with the default one.

### DIFF
--- a/.github/workflows/self-update.yml
+++ b/.github/workflows/self-update.yml
@@ -58,6 +58,20 @@ jobs:
 
                     echo ::set-output name=is_dirty::${IS_DIRTY}
 
+            -   name: Checking for excludes
+                if: success()
+                run: php app/excludes.php
+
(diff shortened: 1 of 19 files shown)


caouecs:
thank you

PR #1607, opened 2021-04-12
[9.x] Improving the project structure

andrey-helldar:
Discussion: https://github.com/Laravel-Lang/lang/discussions/1606


app/main/Concerns/Contains.php
```diff
@@ -8,11 +8,16 @@ trait Contains
 {
     protected function isJson(string $filename): bool
     {
-        return Str::endsWith($filename, '.json');
+        return Str::contains($filename, 'json');
```
caouecs, line 11: Can we use [str_contains](https://www.php.net/manual/en/function.str-contains) directly ?

caouecs (commented):
A question on Str::contains and PHP8 str_contains function

app/main/Concerns/Contains.php
andrey-helldar: Exactly! Forgot about this feature. Thank you!

Yes, we can use it. My package contains this feature for earlier php versions.

I'll update the code in a few minutes.

caouecs:
In TestCase.php
line 64 :  return Str::startsWith($filename, 'validation'); => return str_starts_with($filename, 'validation');

lines 29, 34: return Arr::ksort($content);   I don't find it in [Laravel doc](https://laravel.com/docs/8.x/helpers), and it's ksort(), no ?

andrey-helldar:
> In TestCase.php
> line 64 : return Str::startsWith($filename, 'validation'); => return str_starts_with($filename, 'validation');
> 
> lines 29, 34: return Arr::ksort($content); I don't find it in [Laravel doc](https://laravel.com/docs/8.x/helpers), and it's ksort(), no ?

`Str::startsWith` will be replaced

`Arr::ksort` - this method has never been in the framework. I wrote it myself for my own purposes and use it in several projects. See https://github.com/andrey-helldar/support/blob/main/src/Helpers/Arr.php#L143-L164

app/tests/TestCase.php
```diff
@@ -61,7 +60,7 @@ protected function sort(array &$array): void
 
     protected function isValidation(string $filename): bool
     {
-        return Str::startsWith($filename, 'validation');
+        return str_contains($filename, 'validation');
```
caouecs, line 63: str_starts_with ?

caouecs (changes requested):
Just to be sure

app/tests/TestCase.php
andrey-helldar: Yeah sorry. My mistake. I need to sleep.

andrey-helldar: The code checks the occurrence of words on the file name, so you can use both the `str_starts_with` function and `str_contains`. Taking into account our implementation, changing them is tantamount to saving on matches. We can use both, but I prefer to write correctly right away. The file names `validation.php` and `validation-inline.php` always start with the `validation` word, so it is better to use the `str_starts_with` function.

caouecs:
It's a major version, I'll merge tomorrow, the time to do all versions

andrey-helldar:
https://github.com/andrey-helldar/laravel-lang-publisher/pull/84 is ready to work with version 9 and new structure 😊

andrey-helldar:
Go go go))

caouecs:
done, but we need to update pull requests :)

andrey-helldar:
Yes, I wrote in them that the files were changed.

PR #1615, opened 2021-04-15
update ID lang

bagoesz21:
update ID language


locales/id/validation-inline.php
```diff
@@ -90,15 +90,15 @@
         'numeric' => 'Nilai harus berisi minimal :min.',
         'string'  => 'String harus berisi minimal :min karakter.',
     ],
-    'multiple_of'          => 'The value must be a multiple of :value',
+    'multiple_of'          => 'Isian :attribute harus merupakan kelipatan dari :value.',
```
luisprmat, line 93: keep in mind that the validation-inline.php file does not come with Laravel and the idea of this file is not to put a specific name to each attribute (as in validation.php) but a generic name for the validation attributes. Therefore in the translations of this file the placeholder :attribute should not appear. [View here](https://laravel-lang.github.io/lang/contributing-to-dev.html)

luisprmat (changes requested):

luisprmat:
This is the reason for the failed tests

itsfaqih:
Sorry, I didn't see this pull request and have translated many of them (the json included). I wonder if I still could make the pull request?

locales/id/validation-inline.php
andrey-helldar: image

andrey-helldar:
How to find out exactly what the problem is:

image

image

bagoesz21:
oh, i see, sorry, i'm not focused. Thank you for reminding me.

PR #1616, opened 2021-04-15
[id] Complete Indonesian Translations Update

itsfaqih:

PR #1619, opened 2021-04-15
Added check for ":Attribute" when running tests

andrey-helldar:
I've noticed that some people use `:Attribute` rather than `:attribute` at the beginning of a sentence (for example: [here](https://github.com/Laravel-Lang/lang/pull/1615/files#diff-0ecf1116fcf054e48b4f8a53c508f460bb59310aac195515c515f4f6613661fbL90-L104), [here](https://github.com/Laravel-Lang/lang/pull/1616/files#diff-0ecf1116fcf054e48b4f8a53c508f460bb59310aac195515c515f4f6613661fbL20-L106)). Since the tests are case sensitive, they do not consider `:Attribute` an error. Therefore, I added such a variant of the word spelling when checking the `validation-inline.php` file.

**Before:**

image

**After:**

image

### DIFF
--- a/app/tests/PhpTest.php
+++ b/app/tests/PhpTest.php
@@ -21,7 +21,7 @@ final class PhpTest extends TestCase
                 $this->assertSame(array_keys($source), array_keys($target), $this->messageForLocale($path, $locale));
 
                 if ($this->isInline($filename)) {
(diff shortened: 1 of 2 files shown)


caouecs:
thank you

PR #1620, opened 2021-04-15
Adding a percentage of translations

andrey-helldar:
**Before:**

image

**After:**

image

PS: I just wanted to do 😊

### DIFF
--- a/app/main/Concerns/Countable.php
+++ b/app/main/Concerns/Countable.php
@@ -2,6 +2,7 @@
 
 namespace LaravelLang\Lang\Concerns;
 
+use Helldar\Support\Facades\Helpers\Digit;
 use Helldar\Support\Facades\Helpers\Str;
 use LaravelLang\Lang\Services\Counter;
 
@@ -99,6 +100,6 @@ trait Countable
 
     protected function countFormat(int $count): string
     {
-        return number_format($count, thousands_separator: '\'');
+        return Digit::toShort($count);
     }
 }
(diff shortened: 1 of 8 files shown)


luisprmat:
Idea! How about a percentage for each language?

andrey-helldar:
> Idea! How about a percentage for each language?

This is possible, but on the page of each language - you should not add to the table, otherwise there will be too much text.

But on the pages of localizations, I personally think this information will be redundant, since the number of untranslated elements is already displayed there.

Correct if I'm wrong.

caouecs:
I think it's not interesting because it's just on completed language, not on the number of tags in all languages

andrey-helldar:
@caouecs, are you talking about my proposal or @luisprmat's idea?

caouecs:
> @caouecs, are you talking about my proposal or @luisprmat's idea?

your idea, the global percentage :)

andrey-helldar:
> > @caouecs, are you talking about my proposal or @luisprmat's idea?
> 
> your idea, the global percentage :)

I understood. Now it also seems to me that it is better to calculate by the number of keys, rather than localizations.

I'll add the code shortly.

andrey-helldar:
@caouecs, I updated the code and screenshot in the first post :)

caouecs:
33% is better

andrey-helldar:
> 33% is better

Remove the number of lines and leave only percentages?

caouecs:
No, I said, 33% is better to see than 3% from your first commit ;)

caouecs:
thank you

andrey-helldar:
> No, I said, 33% is better to see than 3% from your first commit ;)

😅

PR #1623, opened 2021-04-16
[de] update translation

sotten:
update german [de] translation


excludes/de.php
```diff
@@ -12,8 +12,108 @@
...
+    'Botswana',
+    'Burkina Faso',
+    'Burundi',
+    'Chile',
+    'China',
     'Code',
+    'Costa Rica',
+    'Curaçao',
     'Dashboard',
+    'Details',
```
andrey-helldar, line 39: `Details` > `Einzelheiten`?

Don't translate this word for sure?
Google Translate says there is a translation.

sotten: Yes I am sure.
Details can be translated with details.
Check "More details" in Google Translate. It translates it with "Mehr Details".
In German we often use the word "Details" for "show me more information about an object"

excludes/de.php
```diff
@@ -12,8 +12,108 @@
...
+    'Haiti',
+    'Honduras',
+    'ID',
+    'Israel',
+    'Japan',
+    'Jersey',
+    'Kiribati',
+    'Kosovo',
+    'Kuwait',
+    'Lens',
```
sotten, line 65: Is "Lens" a geographical place/area/location?
In what context is this term used?  Where is it used? Laravel Nova?

"Lens" in German is "Objektiv" or "Linse"

excludes/de.php
andrey-helldar: Got it. Just asked :)

I know that Google Translate doesn't translate exactly, but this word has a confirmation badge from people.

I think the confirmation is out of context.

image

excludes/de.php
andrey-helldar: @masterix21, can you help with this question?

andrey-helldar: In russian "Lens" is

image

too

andrey-helldar: @sotten, see: https://nova.laravel.com/docs/3.0/lenses/defining-lenses.html

andrey-helldar: It seems to me that they called scopes and repositories with the word "lenses"...

sotten: "Lens" seems to come from Laravel Nova and has nothing to do with geography  ;)
In the Nova documentation this occurs more often.  I think it has something to do with filters or something like that.  I removed Lens from the exclusion list. Maybe it's a Nova proper name too.

andrey-helldar: Now I looked, I [translated](https://github.com/Laravel-Lang/lang/blob/master/locales/ru/ru.json#L283) this word into Russian - "Линза" (en: Lens, de: Linse).

caouecs: the exclusion list is not just for geography

caouecs:
Danke

excludes/de.php
andrey-helldar: Yes, but we couldn't figure out in what context Laravel Nova uses the word "Lens" :)

PR #1625, opened 2021-04-17
Added processors for translating localizations

andrey-helldar:
This code will simply be in the project. It will not run automatically.
I leave it so as not to write it again when needed :)

Now I have a script running that asks Yandex.Translate to translate the missing keys.

When he's finished I'll post a PR for each localization :)

PS: Yes, this is machine translation and some keys need to be corrected. But it's still better than displaying text in English. IMHO.

image

image


caouecs:
thank you

PR #1677, opened 2021-04-17
[de_CH] Machine translation of missing keys

andrey-helldar:

PR #1695, opened 2021-04-18
[de_CH] Machine translation of missing keys | CLEANUP

kohlerdominik:
- Includes #1677 (should be closed therefore)
- Fixed machine translations.


caouecs:
danke

caouecs:
and for de_CH ? https://laravel-lang.github.io/lang/statuses/de-ch.html

I can help if you need to add tags in exclude file

kohlerdominik:
> 
> 
> and for de_CH ? https://laravel-lang.github.io/lang/statuses/de-ch.html
> 
> I can help if you need to add tags in exclude file

I think can copy the exclusion list from de, but I will check in detail later today.

Also, when new translations and exclusions are deteced, do we have any chance that de_CH copies from de rather than from en by default?

PR #1676, opened 2021-04-17
[de] Machine translation of missing keys

andrey-helldar:

PR #1696, opened 2021-04-18
[de] Machine translation of missing keys | CLEANUP

kohlerdominik:
- Includes #1676 (should be closed therefore)
- Fixed machine translations.
- Fixed previously added translations. I guess some kind of machine translation, too, as the grammar is very bad and inconsistent with similar translations.


caouecs:
Danke

caouecs:
Can you check for missing tag in de.json 👼  ? https://laravel-lang.github.io/lang/statuses/de.html

locales/de/de.json
```diff
@@ -430,7 +430,7 @@
     "Restore Resource": "Ressource wiederherstellen",
     "Restore Selected": "Wiederherstellen Ausgewählte",
     "results": "Ergebnisse",
-    "Reunion": "Réunion",
+    "Reunion": "Wiedervereinen",
```
ottenso, line 433: Sorry, but that's completely wrong.  Please also note the context used.  This complete pull request makes it worse than before.

https://de.m.wikipedia.org/wiki/R%C3%A9union

locales/de/de.json
```diff
@@ -280,7 +280,7 @@
     "Leave": "Verlassen",
     "Leave Team": "Team verlassen",
     "Lebanon": "Libanon",
-    "Lens": "Lens",
+    "Lens": "Lupe",
```
ottenso, line 283: Lens is a proper name (Eigenname) of Laravel Nova. Please do not translate.

locales/de/validation-inline.php
```diff
@@ -96,11 +96,11 @@
     'numeric'              => 'Das muss eine Zahl sein.',
     'password'             => 'Das Passwort ist falsch.',
     'present'              => 'Dieses Feld muss vorhanden sein.',
-    'prohibited'           => 'Dieses Feld ist verboten.',
-    'prohibited_if'        => 'Dieses Feld ist verboten, wenn :other gleich :value ist.',
-    'prohibited_unless'    => 'Dieses Feld ist verboten, wenn :other nicht in :values enthalten ist.',
+    'prohibited'           => 'Dieses Feld ist gesperrt.',
```
ottenso, line 99: https://dict.leo.org/englisch-deutsch/prohibit

"gesperrt" has a different meaning. An alternative would be "untersagt", whereby "verboten" is more correct.

locales/de/de.json
caouecs: Here, it's the country, it's not "Réunion" ?

caouecs:
I will fix its today

locales/de/de.json
kohlerdominik: Ok, then I agree its better not to change it

locales/de/de.json
kohlerdominik: Missunderstanding from my side here, totally agree on you

locales/de/validation-inline.php
kohlerdominik: This one was the only one where I could find the PR that introduced it:

> The basic idea of "prohibited" validation rules is that a given field should be prohibited from having data if another field is present or if a field should be allowed in a request at all.
[Source](https://laravel-news.com/laravel-prohibited-validation-rules)

While I'm happy to discuss better wording, I do not think "verboten" is the best explanation for what's going on here. I usually look up the words for translation, IMO it should better match the meaning rather than a hard translation.

kohlerdominik:
> 
> 
> I will fix its today

Hi @caouecs 

I can add the missing translations later today.

I will also fix the two inputs from @ottenso were I agree, and discuss the third one with him were i do not agree with his argumentation.

caouecs:
ok

locales/de/validation-inline.php
sotten: You are right. "verboten" was a bad translation. What do you think of "unzulässig"?

PR #1698, opened 2021-04-19
[km] Update some translations

seakmengc:


caouecs:
thank you

caouecs:
I think "API Token" and "API Tokens" are same in English and in Khmer

but we have just one tag not translated in json file: "The :attribute must be at least :length characters and contain at least one uppercase character and one special character."  Thanks by advance

PR #1700, opened 2021-04-20
[de] Fixes and excludes

kohlerdominik:
As requested in #1696 

@caouecs i'm a little confused about the [status](https://laravel-lang.github.io/lang/statuses/de.html). This keys have translation, but are still shown in the list:

- Action Name
- Action Status
- Iran, Islamic Republic Of
- Isle Of Man
- Lao People's Democratic Republic
- Saint Helena
- Saint Lucia
- Saint Martin 
- Sint Maarten (Dutch part)


locales/de/de.json
```diff
@@ -3,8 +3,6 @@
     "60 Days": "60 Tage",
     "90 Days": "90 Tage",
     ":amount Total": ":amount Gesamt",
-    ":resource Details": ":resource Details",
```
caouecs, line 6: Don't delete tags

caouecs (changes requested):
Don't delete tags

caouecs:
for keys with translation, it's a bug of the todo script

locales/de/de.json
kohlerdominik: i assumed that excludes should be not be keyed. fixed

caouecs:
danke

kohlerdominik:
Ok, [Status ](https://laravel-lang.github.io/lang/statuses/de.html)states now *3 Missing*: 

- `Iran` [is translated](https://github.com/Laravel-Lang/lang/blob/master/locales/de/de.json#L254)
- `Isle Of Man` [is translated](https://github.com/Laravel-Lang/lang/blob/master/locales/de/de.json#L257), only difference in german grammer is that _of_ is lower case. So either add to exclude or consider cases
- `Reunion` [is translated](https://github.com/Laravel-Lang/lang/blob/master/locales/de/de.json#L433), but in german written with an aigu

caouecs:
done

sotten:
I have verified the changes and can confirm them.

PR #1737, opened 2021-05-03
[de] Add translations for new Password Validation Rule

Eixix:
[de] Add translations for new Password Validation Rule

for German


caouecs:
Danke

PR #1758, opened 2021-05-24
[fi] Update translations for Finnish

KJ-L:


andrey-helldar (approved):
Can you also remove the localization code from the machine translation file?

https://github.com/KJ-L/lang/blob/fi-update/app/main/Constants/Machines.php#L18

caouecs:
thank you

PR #1768, opened 2021-05-29
[tr] Completed missing translations

mertasan:


locales/tr/packages/cashier.json
```diff
@@ -7,7 +7,7 @@
     "Extra confirmation is needed to process your payment. Please continue to the payment page by clicking on the button below.": "Ödemenizi işlemek için ek onay gereklidir. Lütfen aşağıdaki butona tıklayarak ödeme sayfasına devam ediniz.",
     "Full name": "Tam adı",
     "Go back": "Geri dön",
-    "Jane Doe": "Jane Doe",
+    "Jane Doe": "Ali Veli",
```
mertasan, line 10: @caouecs 

Ref: https://blog.plover.com/lang/Turkish-John-Doe.html

There were two alternatives. I preferred this. It is more useful because the number of letters is the same.

caouecs:
Thank you

locales/tr/packages/cashier.json
mertasan: @caouecs Sorry. I accidentally read it as "John" Doe. "Ali Veli" is a male name.

If you want to correct it, you can change it to: "Ayşe Fatma".

PR #1778, opened 2021-06-09
fix it translation

trippo:


masterix21:
Grazie @trippo.

@andrey-helldar: PR approved.

caouecs:
Grazie

PR #1789, opened 2021-06-25
[en, es] Remove non ASCII apostrophe from email notification.

luisprmat:
Small changes in Laravel that affect our translations
- https://github.com/laravel/framework/pull/37709/files

The keys that are becoming obsolete should be removed in future versions of this package, we already have several, [among them](https://github.com/Laravel-Lang/lang/blob/master/source/en.json#L9-L11):
- If you’re having trouble clicking the \":actionText\" button, copy and paste the URL below\ninto your web browser:
- Login
- Logout


caouecs:
Gracias

PR #1790, opened 2021-06-26
Fix \/ on it translation

trippo:
Laravel don't recognize translations with \\/ instead /


caouecs:
Hello,
are you sure ? because you are the only person with this problem

masterix21:
@trippo, potresti dirmi in quale occasione ti si presenta il problema? Perché io non lo riscontro in nessuna parte del codice: se mi aiuti a simulare l'errore procediamo con il fix che hai suggerito. Grazie.

@caouecs, I'm digging into the problem with @trippo. I will keep the issue updated. Thanks

trippo:
Sorry i think was the same please ignore

PR #1800, opened 2021-07-05
added translation of one lost sheep in PL ;)

olsza:
I missed one sentence to be translated, ie "Currently Subscribed"


caouecs:
thank you

PR #1803, opened 2021-07-16
 Correction of messages about the problem of clicking on the confirmation button and removal of the corresponding duplicates

theusersimons:
If you look at the [email template](https://github.com/laravel/framework/blob/b1257c4e0d34423ce87a2d60333e547e3ea56631/src/Illuminate/Notifications/resources/views/email.blade.php#L53), you can see that there is a special character `'` in the original message, but in the language resources it was replaced by `’` (apparently a mechanical translation error).

Also, duplicate messages were removed where they were present and now this message is fully localized.


mohamedsabil83:
Many devs still used Laravel 7 or below. For that, we kept it both for now.
It can be removed in the future when the minimum target of this package is Laravel 8.

PR #1827, opened 2021-08-24
[cs] add: Few more Czech translations

maralores:
* Auth translations
* Validation-inline translations
* Exclude init
* Few more CS generic translations


caouecs:
thank you

PR #1835, opened 2021-08-29
[id] Fixed missing translation

elzdave:
1. Fixed missed translation for 'accepted_if' in validation-inline.php
2. Fixed missed translation for 'accepted_if' in validation.php
3. Fixed missed 'The password is incorrect.' and 'Total:' in id.json


caouecs:
Thank you

elzdave:
You're welcome :)

elzdave:
Uhm, I think the 'Total:' should be marked as resolved too, because the 'Total:' in English has the same meaning in Bahasa Indonesia

caouecs:
you can add it in exclude/id.php

PR #1840, opened 2021-09-02
add new `prohibits` validation

sotten:
add `prohibits` to source/validation.php and source/validation-inline.php and translate into German
see https://github.com/laravel/laravel/pull/5681

### DIFF
--- a/source/validation-inline.php
+++ b/source/validation-inline.php
@@ -111,7 +111,7 @@ return [
     'prohibited'           => 'This field is prohibited.',
     'prohibited_if'        => 'This field is prohibited when :other is :value.',
     'prohibited_unless'    => 'This field is prohibited unless :other is in :values.',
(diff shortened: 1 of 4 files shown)


source/validation-inline.php
```diff
@@ -111,6 +111,7 @@
     'prohibited'           => 'This field is prohibited.',
     'prohibited_if'        => 'This field is prohibited when :other is :value.',
     'prohibited_unless'    => 'This field is prohibited unless :other is in :values.',
+    'prohibits'            => 'This field field prohibits :other from being present.',
```
luisprmat, line 114: *"This **field field** prohibits :other from being present."*,  **field** word repeated is right?

sotten: of course not, thanks for the review.

caouecs:
danke

PR #1850, opened 2021-09-14
Update passwords.php in Ukrainian

vnestoruk:
Changed "Provided password recovery code is wrong" to "Password reset link was expired". It's better for end user, because user actually never provide this token (it is hidden) and doesn't even know what the "token" is...


caouecs:
Thank you

PR #1857, opened 2021-09-17
Removes lines not translated or mistranslated

ordago:
Several lines translate characters as *personatges*. That translation is for "the main character in the movie", not letters and numbers.

I will try to get the missing translations translated by a native speaker.


luisprmat:
Why remove untranslated lines instead of translating them?

caouecs:
Hello,
it's not our choice, we prefer to have all lines, because it's easier to translate an existing line, instead of searching the line to translate

PR #1861, opened 2021-09-25
Bug fix for removing keys from `validation.attributes` in translation managers

andrey-helldar:
### Problem

Translation managers don't carry keys from an attribute in an app on update.

### Reason

The transfer is done by reading the keys. There are no `attributes` keys in the `source/validation.php` and `source/validation-inline.php` files.

### Solution

Add keys to the source files.


source/validation.php
```diff
@@ -154,5 +154,39 @@
...
+        'hour'                  => 'Hour',
+        'last_name'             => 'Last Name',
+        'message'               => 'Message',
+        'minute'                => 'Minute',
+        'mobile'                => 'Mobile',
+        'month'                 => 'Month',
+        'name'                  => 'Name',
+        'password'              => 'Password',
+        'password_confirmation' => 'Password Confirmation',
+        'phone'                 => 'Phone',
```
luisprmat, line 180: Could you add the key `photo` that is used in [Laravel Jetstream](https://github.com/laravel/jetstream/blob/2.x/stubs/app/Actions/Fortify/UpdateUserProfileInformation.php#L24)?

luisprmat (changes requested):

source/validation.php
andrey-helldar: Agree. Added.

andrey-helldar: See also https://github.com/Laravel-Lang/lang/pull/1862

caouecs:
I can merge ?

andrey-helldar:
@caouecs, yes

caouecs:
done

andrey-helldar:
Thank you 😊

PR #1895, opened 2021-10-29
Corrected the incorrect translation of the word "Password" for Turkish and added missing Turkish translations.

ugurarici:
Wrong translation was applied for "password" in Turkish. Password is "parola" in Turkish but unfortunately most people confuse this with "Şifre" which means "Cypher" (in encryption). This has been fixed.

In addition, missing translations were added and the Turkish language was completed.


caouecs:
thank you

PR #1899, opened 2021-11-02
[id] Improved language translation

elzdave:
1. Added translation for validation-attributes.php
2. Added translation for validation-inline.php
3. Added translation for validation.php
4. Added excludes for 'Total'


caouecs:
Thank you

PR #1905, opened 2021-11-07
Update pl.json

LocalHeroPro:


locales/pl/pl.json
```diff
@@ -35,7 +35,7 @@
     "All rights reserved.": "Wszelkie prawa zastrzeżone",
     "Already registered?": "Zarejestrowany już?",
     "American Samoa": "Samoa Amerykańskie",
-    "An error occured while uploading the file.": "Wystąpił błąd podczas przesyłania pliku.",
+    "An error occurred while uploading the file.": "Wystąpił błąd podczas przesyłania pliku.",
```
caouecs, line 38: Don't update tags please

caouecs (changes requested):

locales/pl/pl.json
LocalHeroPro: I'm only asking: it's correct with one `r`, or this is typo?
Google Translate see it as typo:
image
Only asking ;]

caouecs: It's not correct but the original file has the error. It's just a tag, so it's not important

LocalHeroPro: So from where that original file comes from? I'll report that to author.

luisprmat: This PR must be sent to Laravel, if they correct it there, we correct us here!

caouecs:
Thank you

PR #1906, opened 2021-11-12
Add missing Dutch translations

RobinBastiaan:
As seen on https://laravel-lang.github.io/lang/statuses/nl.html there are missing translations. I have read this list and added the missing translations where possible.

For some I did not add a translation though, like for example "Korea" as that string is the same as in Dutch. I would like to know if we can safely ignore these strings or if an additional action is needed.


caouecs:
Thank you. Yes we add the tag not translated in exclude directory

luisprmat:
> As seen on https://laravel-lang.github.io/lang/statuses/nl.html there are missing translations. I have read this list and added the missing translations where possible.
> 
> For some I did not add a translation though, like for example "Korea" as that string is the same as in Dutch. I would like to know if we can safely ignore these strings or if an additional action is needed.

https://github.com/Laravel-Lang/lang/blob/master/docs/contributing-to-dev.md#what-should-i-do-if-there-is-a-tag-whose-translation-is-the-same-as-in-english

RobinBastiaan:
Thanks, I will follow up accordingly.

caouecs:
I commit this PR, thank you

you can do a new pull request for exclude tags ;)

PR #1923, opened 2021-12-10
[ID] Fixed missing validation

elzdave:
1. Added missing translation lines for validation-inline.php and
   validation.php


caouecs:
Thank you

PR #1929, opened 2022-01-02
FIX typo in ja

t29kida:
* Delete duplicate line.
* Fix single quotation typo.


caouecs:
Thank you, but it's normal, it's a difference in apostroph

PR #1934, opened 2022-01-10
Replace `may` with `must`

andrey-helldar:
See https://github.com/laravel/laravel/blob/8.x/resources/lang/en/validation.php


caouecs:
thank you

PR #1935, opened 2022-01-10
Removing the `attached` and `relatable` keys

andrey-helldar:
See:
- https://github.com/laravel/laravel/blob/8.x/resources/lang/en/validation.php
- https://github.com/laravel/laravel/blob/7.x/resources/lang/en/validation.php

PS: Tests fail because the composition of the keys has been changed


andrey-helldar:
Stop, it's Nova :)

image

PR #2014, opened 2022-01-27
Updated Brazilian portuguese translation with missing Translations

EuCarlos:
In line [293](https://github.com/Laravel-Lang/lang/blob/master/locales/pt_BR/pt_BR.json#L293) there is a key "Jane Doe" which can be translated as "Indigente", "Maria nínguem" and "Desconhecida" but it is best to keep it as "Jane Doe".

In line [636](https://github.com/Laravel-Lang/lang/blob/master/locales/pt_BR/pt_BR.json#L636) it is already correct.

pt_BR.json is finished.


andrey-helldar (approved):
Obrigado

PR #2077, opened 2022-03-20
[de] Missing German translation

sotten:
Translate "This coupon code can only be used by new customers."


andrey-helldar:
Dankeschön

PR #2106, opened 2022-04-05
[de] part update after new Nova 4 translation keys

sotten:
some parts are now updated according to new Nova 4 translation keys


andrey-helldar:
Danke

PR #2139, opened 2022-04-16
[fr] Fix malformed french translations

shadoWalker89:
The Following french translations are malformed, this PR fixes that


caouecs:
Les deux versions se valent, mais merci

PR #2140, opened 2022-04-16
[fr] Add equality to the sentence

shadoWalker89:
The updated translation for 'min' was saying `File size should be greater then :min` This PR adds `ou égale` which means `or equal`


caouecs:
Merci, il faut penser à modifier sur validation-inline.php aussi

shadoWalker89:
C'est fait https://github.com/Laravel-Lang/lang/pull/2145

PR #2145, opened 2022-04-18
[fr] Add equality to the sentence

shadoWalker89:
The updated translation for 'min' was saying File size should be greater then :min This PR adds ou égale which means or equal


caouecs:
Merci

PR #2172, opened 2022-05-09
Update password field

Khuthaily:
The value of `password` is now an array instead of a string.


andrey-helldar:
In the current state of the project, you cannot change its structure. The project contains translations since Laravel 5.8. After my vacation, I will change the project structure to support the new project structure.

andrey-helldar:
Added in version 11.0. Thank you!

PR #2177, opened 2022-05-12
Migrate password rules from json to validation(-inline).php

luisprmat:
This PR allows to migrate _password validation rules_ from `{locale}.json` to `validation.php` in the array *password*.
Translations found in the `{locale}.json` were moved to `validation.php`


andrey-helldar:
The current release contains translations since Laravel 5.8. Therefore, we cannot simply take and change the structure of existing keys.

I [started](https://github.com/andrey-helldar/Laravel-lang/tree/11.x) to develop code to change the structure of the project, but I will complete it after the vacation.

PR #2185, opened 2022-05-20
Update zh_TW.json

whsyu:
Google uses "個人資料" for "Profile", and I think it's ok to align with them.
remove unnecessary "此/這個" for some "This-/The-"


whsyu:
opps...  will create a new one

PR #2194, opened 2022-05-25
[fr] Update missing fr translations

vodkhard:
update fr translations referenced in https://laravel-lang.com/statuses/fr.html


caouecs:
Pour moi, c'est bon, merci

For me, it's ok, thank you

PR #2208, opened 2022-06-12
Correct the Cameroon country name in portuguese

ravelinodecastro:
"Camarão" in english means shrimp, the correct name of Cameroon in portuguese is Camarões.


jorgercosta (approved):
That particular change is ok but the file needs a full review, multiple countries are wrong

andrey-helldar:
Obrigado

PR #2224, opened 2022-06-20
Update validation-inline.php

nunhes:
Update translation


andrey-helldar (changes requested):
Correct the codestyle of the file. The project has a `.editorconfig` file with a set of rules for the IDE. Use the EditorConfig plugin and apply code auto-formatting.

andrey-helldar (changes requested):
The inline translation file must not contain references to field names.

For example,

Main file:
> The :attribute must be accepted.

Inline:
> This field must be accepted.

PR #2249, opened 2022-06-30
fix fa translation

ariaieboy:


ariaieboy:
> I probably took the PR too quickly. Sorry :)
> 
> PS: You don't need to change `:Attribute` to `:attribute`. Laravel has a feature out of the box where `:Attribute` capitalizes the first letter after a replacement. Since the sentence begins with a capital letter, it is correct to do so. In the case of localization fa, this does not play any role.

in my repo the PHPUnit test failed because of that "A" that's why I've changed it.

andrey-helldar:
Hmm... I'll check it out :)

andrey-helldar:
Can I accept PR or will you still work?

ariaieboy:
> Can I accept PR or will you still work?

I think it's all good.

about the test that was my fault I didn't read the error message correctly that was because of using :atteribute in inline json. :)

andrey-helldar:
Yes, I'm confused too)

PR #2256, opened 2022-07-03
[ja] update translation

wadakatu:
added some translations in ja files.

Hope it helps.


wadakatu:
I got a weird failure on phpunit result.

```
There was 1 failure:
1) Tests\PluginTest::testInline
The locales/de/json-inline.json file must not contain the value: :Attribute
:Attributeは有効なロールである必要があります。
Failed asserting that true is false.
/home/runner/work/lang/lang/vendor/laravel-lang/status-generator/tests-app/Concerns/Assert.php:31
/home/runner/work/lang/lang/vendor/laravel-lang/status-generator/tests-app/Concerns/Assert.php:14
/home/runner/work/lang/lang/vendor/laravel-lang/status-generator/tests-app/TestCase.php:37
```

I did not edit any file in `de` directory.
Should I remove all `:attribute` in `ja/json-line.json` to solve this failure ?

andrey-helldar:
Localization error in tests fixed.

Screenshot_2022-07-03-19-21-27-751_com android chrome

andrey-helldar:
Yes, for this test to pass, no value in the inline file must refer to an attribute name.

andrey-helldar:
For example:

```json
"The given :attribute has appeared in a data leak. Please choose a different :attribute.": "The given field has appeared in a data leak. Please choose a different field."
`

**Comment by wadakatu** [comment]
@andrey-helldar 

sorry for late reply.
I removed all `:attribute` in json-inline.json, and passed phpunit.
Thank you for your support.

File: locales/ja/json.json [hunk 1]
```diff
@@ -20,7 +20,7 @@
     "Actions": "アクション",
     "Add": "追加",
     "Add a new team member to your team, allowing them to collaborate with you.": "新しいチームメンバーを追加",
-    "Add additional security to your account using two factor authentication.": "セキュリティ強化のため二段階認証を追加",
+    "Add additional security to your account using two factor authentication.": "セキュリティ強化のため二要素認証を追加",
```
andrey-helldar, line 23:
```suggestion
    "Add additional security to your account using two factor authentication.": "セキュリティ強化のため二要素認証を追加。",
```

locales/ja/php.json
```diff
@@ -28,7 +28,7 @@
     "digits_between": ":Attributeは、:min桁から:max桁にしてください。",
     "dimensions": ":Attributeの画像サイズが無効です",
     "distinct": ":Attributeの値が重複しています。",
-    "doesnt_start_with": "The :attribute may not start with one of the following: :values.",
+    "doesnt_start_with": ":attributeは次のうちいずれかで始まってはいけません。: :values",
```
andrey-helldar, line 31: Is it correct that the sentence ends with a full stop before a colon with a choice of options?

Maybe it should be like this? Or am I wrong?

```suggestion
    "doesnt_start_with": ":attributeは次のうちいずれかで始まってはいけません: :values。",
```

andrey-helldar (changes requested):

locales/ja/json.json
wadakatu: @andrey-helldar 

Thank you for your review.
I thought there are several similar sentences that there is no `。` at end of the sentence.
Do you think that I have to fix all of them as you suggested?

andrey-helldar: We determine the presence of full stops at the end of sentences by the source phrase for translation.

Since the `Add additional security to your account using two factor authentication.` clause ends with a full stop, we also use it.

The fact that in the previous version of the translation there was no point, this is my omission. Apparently, I missed this point during the code review.

In general, we try to stick to the same style of writing as in the original.

wadakatu: @andrey-helldar 

I understood.
I am going to fix the it as soon as possible.

locales/ja/php.json
wadakatu: @andrey-helldar 

I am not 100% sure which one is correct.
it might be my sentence. it might be your sentence. It could be both.

is `:values` something like this?
`A, B, C`

locales/ja/json.json
andrey-helldar: If it helps, GitHub provides a quick way to commit changes as long as they're valid.

image

locales/ja/php.json
andrey-helldar: Given the name of the key, under the hood the validator will explicitly check with the `in_array` function, so the final sentence in English might look like this:

> The :attribute may not start with one of the following: :values.
> The first name may not start with one of the following: foo, bar, baz.

When first name is `fooable`, for example.

I don't know the Japanese grammar rules, but in many countries the full stop is placed at the end of a sentence, not before the list of meanings.

That's why I checked if it's correct.

locales/ja/json.json
wadakatu: Wow, that's cool feature that I've never used before.
Thank you for your information.
I appreciated it :)

locales/ja/php.json
andrey-helldar: We stick to the writing format, but the end goal is grammatically correct sentences.

andrey-helldar (approved):
Thank you

PR #2274, opened 2022-07-21
[es]: Update json-inline

luisprmat:


andrey-helldar (approved):
Among other things, yesterday I fixed the [`automatic update`](https://github.com/Laravel-Lang/lang/tree/main/locales/en) of the English localization when changing the source files.

The exception file contains values, not keys.

Thus, the value "Macedonia" does not exist in the source files, because the error corrected the spelling, and in English localization this key is converted to [`North Macedonia`](https://github.com/Laravel-Lang/lang/blob/main/locales/en/json.json#L350).

Therefore, you don't need to add it to the exclusion file.

image

I will accept this PR, and the script will automatically remove the extra word when merging 🙂

Gracias 🤝

andrey-helldar:
I will fix this bug soon:

image

PR #2287, opened 2022-08-02
Swahili translation

Pheogrammer:
I have translated the added keywords and modified the inline translation as needed to be done


andrey-helldar (approved):
Thank you!

PR #2329, opened 2022-08-16
[sw]: Translation of added keyword(s)

Pheogrammer:
Translation of added keyword: "The provided two factor recovery code was invalid."


andrey-helldar (approved):
Ahsante

PR #2401, opened 2022-09-26
replace field in arabic language by :attribute to insure the dynamic …

averooos:
I have replaced الحقل by :attribute because I feel it more accurate in dynamic way than static


andrey-helldar:
Inline translations must not contain a mention of the field name.

PR #2457, opened 2022-11-11
[ro]: Translation of added keyword on ro/php-inline.json

Van4kk:
- translation of added keyword `lowercase`
- some grammar fixes


andrey-helldar (approved):
Mulțumim

PR #2468, opened 2022-11-15
[sw]: Translation of added keywords 

Pheogrammer:
I have translated the added keyword in php-inline and php files


andrey-helldar (approved):
Ahsante

PR #2470, opened 2022-11-15
[sw]: Translation of added keywords (for added keywords from 9.x keys)

Pheogrammer:
I have translated the newly added keywords from #2467


andrey-helldar (approved):
Ahsante

PR #2493, opened 2022-11-22
[it]: Update translations

masterix21:


locales/it/json.json
```diff
@@ -686,7 +686,7 @@
     "This user has already been invited to the team.": "Questo utente è stato già invitato nel team.",
     "Timor-Leste": "Timor-Leste",
     "to": "a",
-    "To": "To",
+    "To": "a",
```
andrey-helldar, line 689: Is it correct here that "To" would be translated as "a" and not "A"?

masterix21:
My mistake, sorry. Thanks for your effort.

andrey-helldar (approved):
Grazie

PR #2543, opened 2022-12-21
[pl]: Decimal key added

makowskid:


locales/pl/php-inline.json
```diff
@@ -21,7 +21,7 @@
     "date": "Pole nie jest prawidłową datą.",
     "date_equals": "Pole musi być datą równą :date.",
     "date_format": "Pole nie jest w formacie :format.",
-    "decimal": "This field must have :decimal decimal places.",
+    "decimal": "Pole :attribute musi mieć :decimal miejsc po przecinku.",
```
andrey-helldar, line 24: An inline translation must not refer to an attribute name.

andrey-helldar (changes requested):

locales/pl/php-inline.json
makowskid: @andrey-helldar thanks for pointing this out, fat fingers, sorry - fixed already

andrey-helldar (approved):
Dziękuję

PR #2570, opened 2023-01-03
[pl]: New keys updated

makowskid:


makowskid:
Hi

I've noticed an issue in all translations. There are 2 similar keys:
1. Reunion
2. Réunion

I think they both are related to https://pl.wikipedia.org/wiki/Reunion
but in most translations, the first key is directly translated from its English meaning, which I think might be an error and both should be always just excluded from any translations.

andrey-helldar (approved):
Dziękuję

andrey-helldar:
Do you mean that you need to replace the country name in all localizations?

PR #2575, opened 2023-01-06
[fr]: use of the typographic apostrophe ’ instead of '

forxer:


caouecs:
Hi,
I'm against the apostrophe ’, ' is more universal

WhereIsLucas:
Hi,
Thanks for the PR. 
As @caouecs, I prefer the `'` caracter.

forxer:
The apostrophe in the "single quote" form is just a legacy of computing which had a restricted character set. Now that UTF-8 is the de-facto standard I see no reason to deprive ourselves of the true typographic apostrophe.

Especially since we are on laravel-lang, this is where we should implement the linguistic particularities of the different languages :)

This is even more particularly true for quotation marks, which are written in different ways depending on the languageand the context:

- “…”
- « ... »
- “ ... ” 
-  ‘ ... ’ 
-  ‹ ... › 
- '  ... '
- etc.

andrey-helldar:
And this is where I don't quite agree. For example, in Russian, both the variant `"..."` and `«...»` can be used to frame a sentence in quotation marks. The variant `«...»` is used in printed materials, but `"..."` - in the network. Yes, you can see sites using `«...»` for framing, but this is rare. The most commonly used, let's say, monospace-like options, that is, `"..."`.

I can assume that something similar happens in the case of the apostrophe in French grammar.

But you gave me an idea on how to expand the functionality of the [publisher](https://publisher.laravel-lang.com).

forxer:
Yes it is clearly a difference between typographic version and computer version. And this for historical reasons of the beginning of computing and its technical constraints at the time. There is also the history upstream of typewriters...

But now that we no longer have the technical constraints, I find it a shame not to make full use of all the tools at our disposal.

A great example of this is [League Commonmark's "Smart punctuation" extension](https://commonmark.thephpleague.com/2.3/extensions/smart-punctuation/).

andrey-helldar:
@forxer, [publisher version 14.5](https://github.com/Laravel-Lang/publisher/releases/tag/v14.5.0) added the ability to apply smart punctuation. You can activate it in the settings of your application and get the desired result when updating.

See more: https://publisher.laravel-lang.com/features/smart-punctuation.html

forxer:
@andrey-helldar pefect, thank you.

PR #2582, opened 2023-01-10
[es]: Update php.json adding spaces

luisprmat:


andrey-helldar (approved):
I don't know if this was done on purpose or not, but so be it.

Grazie!

PR #2601, opened 2023-01-19
[fr]: Updated translations

bcag2:
added some _exclude tems (mainly USA areas) and added some current translations.
I also replace `(e)` feminine by `·e`, for example : male writing of english "sure" word is in french "sûr" and female "sûre". It was writing "sûr(e)", I replaced by "sûr·e" which is a more and more using inclusive form.


caouecs:
I like this way, I agree to merge

but attention with another ticket on French version,

PR #2810, opened 2023-02-11
[es]: Standardise "must" and "field" in translations

luisprmat:
This PR is related to the proposed changes in:
https://github.com/laravel/framework/pull/45974 and
https://github.com/laravel/framework/pull/45989
for Laravel 10.

Our system does not have a way to notify us if there were changes in the translations of existing keys of `.php` based files (It only informs us if there are new keys).

I don't know if this will have a big impact on translations from other languages, but it would be nice if the @Laravel-Lang/laravel-lang  team would review it.


andrey-helldar (approved):
Gracias

PR #2926, opened 2023-02-22
[es]: Translation error

joc-luis:
Replace 'The' with 'El' on 'in' message

### DIFF
--- a/locales/es/php.json
+++ b/locales/es/php.json
@@ -58,7 +58,7 @@
     "gte.string": "El campo :attribute debe tener como mínimo :value caracteres.",
     "if_you_delete_team_all_data_will_be_deleted": "Si usted elige borrar el equipo, todos los datos se eliminarán de manera permanente.",
     "image": "El campo :attribute debe ser una imagen.",
-    "in": "The :attribute seleccionado no es válido.",
+    "in": "El :attribute seleccionado no es válido.",
     "in_array": "El campo :attribute debe existir en :other.",
     "integer": "El campo :attribute debe ser un número entero.",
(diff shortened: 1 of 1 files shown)


locales/es/php.json
```diff
@@ -58,7 +58,7 @@
     "gte.string": "El campo :attribute debe tener como mínimo :value caracteres.",
     "if_you_delete_team_all_data_will_be_deleted": "Si usted elige borrar el equipo, todos los datos se eliminarán de manera permanente.",
     "image": "El campo :attribute debe ser una imagen.",
-    "in": "The :attribute seleccionado no es válido.",
+    "in": "El :attribute seleccionado no es válido.",
```
andrey-helldar, line 61: Other translations begin with the words "El campo". Maybe it's needed here too.

andrey-helldar (changes requested):

locales/es/php.json
luisprmat: In the source [`'in' => 'The selected :attribute is invalid.'`](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Translation/lang/en/validation.php#L69) doesn't appear the word *field* whose meaning is *campo*.

This translation doesn't require the word "*campo*". 
It was an error in my translation so the observation as proposed is correct.

luisprmat (approved):

luisprmat:
Gracias por la correción @joc-luis.

locales/es/php.json
andrey-helldar: OK

andrey-helldar (approved):
Gracias

PR #3267, opened 2023-05-07
[nb]: Update translations (Norsk Bokmål)

dsbilling:


andrey-helldar (changes requested):
You need to sort the values in the exclusion file alphabetically.

dsbilling:
@andrey-helldar Fixed, but now there is an other issue with the other workflow :)

PR #3289, opened 2023-06-19
[nl]: Translated new keys

chillbram:
First validation key which references the reader with "je", but I think it's necessary as "onbevoegde/ongeautoriseerde waarde" would not scan as normal Dutch and would lower comprehensibility.


andrey-helldar (approved):
Dankjewel

PR #3603, opened 2023-09-13
[ro]: Updated translations

JakyeRU:


andrey-helldar:
@Van4kk, can you check this one?

andrey-helldar (approved):
I think we can accept these changes and adjust them after release in case something is incorrect.

PR #3625, opened 2023-11-03
fixing broken automated translations of parameters

makowskid:
I've noticed that some earlier automatic processing has messed up a lot of translation files.
Case-sensitive matters in these parameters a lot. (Even in my projects I can see these broken parameter strings are not replaced with proper values).

Is there some script that is being run periodically that breaks those?


makowskid:
Ned to review it again - because in some cases capitalization of first letter makes sense

PR #3708, opened 2023-11-09
[sw:] Translation of added keywords

Pheogrammer:
I have translated the added keywords in php-inline and php


andrey-helldar (approved):
Ahsante

PR #4163, opened 2024-01-15
[th]: Update json.json: "Thailand": "ประเทศไทย" -> "ไทย"

bact:
Change "Thailand" translation from "ประเทศไทย" to "ไทย".

Based on the translation of other country names, like:

```json
    "Tajikistan": "ทาจีกิสถาน",
    "Tanzania": "แทนซาเนีย",
    "Tanzania, United Republic of": "แทนซาเนีย, สหสาธารณรัฐ",
```

They don't include the prefix "ประเทศ" ("country") in their translation.

So the translation of "Thailand" should not include "ประเทศ" as well.

This will make it consistent for the user and make it easier to find country names when sorted alphabetically (with a locale-aware collator).

### DIFF
--- a/locales/th/json.json
+++ b/locales/th/json.json
@@ -628,7 +628,7 @@
     "Subscription Information": "ข้อมูลการสมัครสมาชิก",
     "Subscription Pending": "การสมัครสมาชิก รอดำเนินการ",
     "Sudan": "ซูดาน",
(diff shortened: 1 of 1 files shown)


andrey-helldar (approved):
ขอบคุณ

PR #4165, opened 2024-01-15
[th]: Update php-inline.json - remove fullstops

bact:
- Remove a fullstop at the end of sentence, they are not a symbol in standard Thai writing
- Retranslate `alpha`, `alpha_dash`, and `alpha_num`: "letter" here means something similar to an alphabet ("ตัวอักษร") and not a thing inside an envelop ("จดหมาย") as it is currently get translated.


andrey-helldar (approved):
ขอบคุณ

andrey-helldar:
We also have a translation status page. It helps when you need to find untranslated meanings.

https://laravel-lang.com/statuses.html

PR #4166, opened 2024-01-16
[th]: Update json-inline.json

bact:
Make the translation of the variations of a similar sentence (on letter/number/symbol combination requirements) looks consistent across.


andrey-helldar (approved):
ขอบคุณ

PR #4298, opened 2024-02-19
[de]: Improve translation

es-ka:
The strings "Showing" and "results" are changed to be correct in pagination context:
[src/Illuminate/Pagination/resources/views/bootstrap-5.blade.php](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Pagination/resources/views/bootstrap-5.blade.php)


sotten (approved):

andrey-helldar (approved):
Danke!

PR #4301, opened 2024-02-21
[es]: Update json.json

luisprmat:
This translation often describes an action but **Vista** is not an action!


andrey-helldar (approved):
Gracias

PR #4302, opened 2024-02-27
[de]: Fix password.sent translation

oliworx:
"Passworterinnerung" is wrong, it means password remember and not password reset


sotten (approved):

andrey-helldar (approved):
Danke

PR #4305, opened 2024-03-04
[th]: Update translations of set up key and recovery key

bact:
Make all of them consistent to each other


andrey-helldar (approved):
ขอบคุณ

PR #4432, opened 2024-03-15
[ja]: Inconsistent use of words for `numeric`.

tscontenna:
`numeric` validation rule accepts negative float value like `-1.3`, `0.5E-3` etc.

The word `数字` implies that some characters like `-`, `.` are not included.


wadakatu (approved):
@tscontenna 
LGTM 👍
修正ありがとうございます！

andrey-helldar (approved):
ありがとうございます

PR #4555, opened 2024-03-19
[source]: Remove support for Laravel Framework 10, Lumen Framework, Breeze 1, Cashier Stripe 13, 14, Jetstream 2, 3, 4

andrey-helldar:
Removed support for outdated versions of projects.

- [x] Update all localization files by adding values to the `(and :count more errors)` key.
- [x] Supplement the regular expression in `Status Generator` to search for the text `(and :count more errors)`.

PS: Do not measure until the "Status Generator" functionality is extended to be compatible with the `trans_choice` function.


andrey-helldar:
@Laravel-Lang/laravel-lang, hey!

So now telling you what just happened and why a new major version of the Lang project was released.

The thing is that Laravel 11 disabled the translation for the error count phrase returned in the main validation message. Because of this, I released version 14.5 with support for Laravel 11, Breeze 2, Jetstream 5, and Cashier Stripe 15.

But then Laravel 11.0.7 version adopted my PR (https://github.com/laravel/framework/pull/50546) to improve translation for this field. It now supports multiple forms and is handled by the `trans_choice` method (https://github.com/laravel/framework/pull/50560).

And this became a problem because our mechanism goes through all files and collects values by key name, and since the name is the same, `array_merge` applies the last one taken. That's why the translation of the phrase became incorrect - always in the same plural form.

Since we jointly decided that we would support our projects for major versions of the Laravel framework as long as they are officially supported (major support for Laravel 10 ends on [August 6, 2024](https://laravel.com/docs/11.x/releases#support-policy)), I followed three basic steps in the `Lang` project:

1. In version 14, turned off language file updates for Laravel 11 only (all other packages are available);
2. Released a new, 15th version of the Lang project, removing Laravel 10 and all packages that only worked with it.
3. Added support for version 15 to the `Common` package, since technically it doesn't conflict with it in any way.

This way we can continue to support version 10 of the framework (personally, I doubt that new translations will be actively added to it now, so we don't have to worry), and version 11. The only difference is that there will be two releases for each version if necessary.

Those using `Common` 6 and Laravel 11 versions will simply run the console command `composer update` and the package will update itself to version 15 without any issues. That's awesome!

That's all :)

https://laravel-lang.com/release-notes.html#lang

mohamedsabil83:
Good idea and good works 💯

andrey-helldar:
@mohamedsabil83, thank you 😊

PR #4743, opened 2024-05-01
[nl]: json.json fixes

chillbram:
I did a pass through all translations in nl/json.json and fixed a bunch of AI translation errors. The translations regarding Resources are still scuffed I think, but I don't use Nova(?) so don't know what a fitting translation would be for those keys


andrey-helldar:
I don't use Nova either, so we'll wait for the feedback post facto :)

andrey-helldar (approved):
Dankjewel

PR #4805, opened 2024-05-15
[el]: Translated new "contains" key and missing inline translations

michaelkonstantinou:
This pull request updates the following:

- New "contains" key
- PHP inline values that were not translated
- A few typos


andrey-helldar (approved):
Ευχαριστούμε

PR #4849, opened 2024-07-12
[el]: Updated Island names and added new keys

michaelkonstantinou:
Translations for the Greek language:
- Updated several island names that were partially or not translated
- Added translations for new keys


andrey-helldar (approved):
Ευχαριστούμε

PR #4915, opened 2024-09-24
[lt]: Update json.json

mantax559:


andrey-helldar (approved):
For the future, those changes can be commit and push in a previous branch so that they go into that PR. In this case I accept them both, so they will complement each other.

Dėkoju!

PR #4934, opened 2024-11-04
[de_CH]: Synced with [de]

kohlerdominik:


kohlerdominik:
@andrey-helldar Is there any existing mechanic, so we could automatically sync [de_CH] from [de] for missing keys? de_CH is very similar to de, and fits out of the box easily.

There is one conversion that could be auto-applied:
- `ß` does not exist in de_CH, and should be replaced with `ss`. So search-replace `ß`→`ss`.

As soon as before-values in de and de_CH are not equal, they should no longer be auto-synced, obviously.

andrey-helldar:
There are no mechanics for synchronizing these languages because, despite their similarities, they have completely different localization codes.

To replace `ß` with `ss`, you can use file-by-file text replacement in the IDE (Ctrl+H).

andrey-helldar:
@sotten hey! Can you check another one?

kohlerdominik:
> There are no mechanics for synchronizing these languages because, despite their similarities, they have completely different localization codes.
> 
> To replace `ß` with `ss`, you can use file-by-file text replacement in the IDE (Ctrl+H).

Hi
Yes, this is what I do right now. But it requires someone to do it manually; so if there would have been something similar in place, we could have automated it. But I agree that it's not worth to do it for one language, hence the question.

andrey-helldar (approved):
Danke!

PR #4937, opened 2024-11-08
[en]: Updated incorrect inline strings

kohlerdominik:
Seems like inline strings were automatically updated at some point. But a lot of values don't make sense with the applied logic.
Example:
`The [Begin Date] must be a date before :date.` was translated to inline `The field must be a date before :date.`, but it should be `The field must contain a date before :date`.

So because a lot of placeholders refer to the value and not the key, if the field should be referenced, most strings need to be updated for something like replacing `be` to `contain`: 

Simplified example: `Value should be X` → `Field should contain X`

---

This PR updates all inline strings affected.


andrey-helldar:
There is no point in touching the English localization files, as they are updated automatically.

Even if you accept this PR, a script will immediately run to check them and fix everything.

andrey-helldar:
Technically we can turn off the automatic update of English localization, but we should only do this if it is guaranteed to be wrong and only a manual update will fix the situation.

Next, let's take the key `The :attribute must be at least :length characters and contain at least one number.`. You suggest using `The string must be...` instead of `This field must be...`. In this case I agree, because these keys are related to password validation, and it can only be a string.

But changing the phrase `The given :attribute has appeared in a data leak.` to `The given string has...` can be misleading, because the field may not be a string, but a number, for example.

There's also a suggested translation for `“dimensions”: “This image has invalid dimensions."`, which could also refer to video files, which would already be an error. Here I would consider what can be done.

Nevertheless, overall, your suggestion looks like a real improvement. The only thing is that I personally don't know English and use online translators, so I can't fully say what is right and what is wrong. In this regard, I urge @Laravel-Lang/laravel-lang. Who knows English well? What do you think?

kohlerdominik:
Hi @andrey-helldar

Thanks for the real good feedback. I'm not native english speaker either. I tried my best, but I'm sure there's quite some room for improvement. And I'm very glad if someone very proficent in the english language could check this again. 

Also, good point about the `dimensions`, there are probably a few other "edge cases" aswell.

However, I feel like with having a majority of sentences usable, it's way more likely to get some native speaker to use this package and therefore contributing their knowledge. Same for edge cases.

andrey-helldar:
I agree. It remains to wait for those who know English well :)

@Laravel-Lang/laravel-lang hi! Can you help us?

andrey-helldar:
> A person would say “This text must be…”

I'll put my 2 cents in here too :)

If for a string we can still say “This text should be”, then, for example, for a field with a password it will be unnatural, and in this case it is more correct to write “This password should be...”. Or the name of the product, for example - “This product should be...”.

PR #6108, opened 2025-02-20
[fr]: Update translations

squiaios:


andrey-helldar (changes requested):
You don't need to change the file format, especially to an incorrect one.

WhereIsLucas (approved):
Translation is ok for me

PR #6130, opened 2025-02-28
[ko]: Translate the word `View` more correctly.

cable8mm:
The English word "view" can function as both a verb and a noun, but in Korean, these meanings should be distinguished.

I suggest the following change:

```diff
- "View": "뷰",
+ "View": "보기",
```

In Korean, "뷰" is commonly used to refer to a sight, landscape, or scene, rather than the act of looking or seeing. To better reflect the intended meaning, "보기" would be a more appropriate translation.

view_cannot_be_understood

I believe no Korean would understand the image above - "뷰".

Reference link - https://nova.laravel.com/docs/v4/resources/fields#tag-field

### DIFF
--- a/locales/ko/json.json
+++ b/locales/ko/json.json
@@ -801,7 +801,7 @@
     "Verify Email Address": "이메일 주소 확인",
     "Verify Your Email Address": "이메일 주소 확인",
     "Viet Nam": "베트남",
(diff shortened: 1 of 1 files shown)


andrey-helldar (approved):
감사합니다

PR #7245, opened 2025-06-04
[de]: Translated missing strings

hebbet:


andrey-helldar:
@sotten Hi! Can you check this one?

PR #7278, opened 2025-08-07
[ja]: Fix Typo in Japanese

zigzagdev:
## what i done

``` before
"Mali": "小さい"
```
```after
"Mali": "マリ"
```

I made this change because, judging from the surrounding context, it seems that "Mali" is referring to the country rather than the adjective meaning "small" in Croatian or other Slavic languages.

However, if this key was intentionally translated to mean "small" in another language, please feel free to close this PR. I would appreciate your clarification in that case.

Thank you for your time and consideration.


wadakatu (approved):
LGTM!

PR #7279, opened 2025-08-07
[bg]: Fix string parameter not being displayed

dyanakiev:
The parameter in the translated string was actiontext instead of actionText.

CleanShot 2025-08-07 at 09 44 25


andrey-helldar (approved):
Благодаря ти

PR #8279, opened 2026-04-25
updated missing translations

TheXERC:


andrey-helldar:
These files do not need to be changed manually.

PR #8285, opened 2026-05-08
[ja]: Add missing Japanese translation for encoding rule

wadakatu:
## Summary

- Added missing Japanese translations for the `encoding` validation rule in `php.json` and `php-inline.json`
- This rule was already translated in other locales but was left untranslated in ja, so I translated it to maintain consistency

## Changes

- `php.json` `encoding`: "The :attribute field must be encoded in :encoding." → ":attributeは、:encodingでエンコードされている必要があります。"
- `php-inline.json` `encoding`: "This field must be encoded in :encoding." → ":encodingでエンコードされている必要があります。"

Hope it helps.


andrey-helldar (approved):
ありがとうございます

PR #8301, opened 2026-05-26
[source]: Fix version compare for laravel/passkeys

luisprmat:
Publisher don't recognize dev version because laravel/passkeys now uses tagged versions


andrey-helldar (approved):
Gracias

PR #8335, opened 2026-07-25
[nl]: Fix untranslated base64/encoding validation strings

arjankapteijn:
## Summary

The Dutch (`nl`) translation for the `base64` and `encoding` validation rule messages (added for Laravel's newer `base64` and `encoding` validation rules) was identical to the English source text — i.e. never actually translated, just copied verbatim. Flagged as missing by the [`nl` translation status page](https://github.com/Laravel-Lang/lang/blob/main/docs/statuses/nl.md).

| File | Key | Old (untranslated) value | New value |
|---|---|---|---|
| `locales/nl/php.json` | `base64` | `The :attribute field must be a valid Base64 string.` | `:Attribute moet een geldige Base64-string zijn.` |
| `locales/nl/php.json` | `encoding` | `The :attribute field must be encoded in :encoding.` | `:Attribute moet gecodeerd zijn in :encoding.` |
| `locales/nl/php-inline.json` | `base64` | `This field must be a valid Base64 string.` | `Dit veld moet een geldige Base64-string zijn.` |
| `locales/nl/php-inline.json` | `encoding` | `This field must be encoded in :encoding.` | `Dit veld moet gecodeerd zijn in :encoding.` |

Only the values were changed, not the keys. Wording follows the existing style conventions in these files, matching the neighboring `json`/`hex_color`/`ip`/`url`/`uuid` entries (e.g. `"json": ":Attribute moet een geldige JSON-string zijn."`).

## Test plan

- [x] `locales/nl/php.json` and `locales/nl/php-inline.json` remain valid JSON
- [x] Only the 4 affected values were changed; no keys were added, removed, or renamed


andrey-helldar (approved):
Dankjewel

PR #8365, opened 2026-08-15
[pl]: update polish translation

Eurynomosek:
First PR in my live, i hope its ok.


andrey-helldar:
The Lang project contains only the keys officially included in Laravel products.

The set of keys is synchronized automatically.

You can add new keys, but they will be removed immediately after the PR is merged.

PR #8370, opened 2026-08-31
[fr]: Translated missing value and fixed punctuation

TimeglitchD:
Added a missing translation for encoding and added a period to uuid, like the other fields have.


andrey-helldar (approved):
Merci

PR #8374, opened 2026-09-01
[hu,ko,lt,sl,sq,sr_Cyrl,sr_Latn,sr_Latn_ME,tr]: Restore corrupted placeholders

katerynaKhar:
Laravel substitutes with `strtr()` against `:name`, `:Name` and `:NAME`. None of
these nine values matches any of those three forms, so the token is printed
literally and the value never reaches the user.

| locale | key | was | cause |
|---|---|---|---|
| `hu` | `Load :perPage More` | `:perpage` | case |
| `lt` | `Load :perPage More` | `:perpage` | case |
| `tr` | `Load :perPage More` | `:Perpage` | case |
| `ko` | `Managing billing for :billableName` | `:BilableName` | typo, one letter |
| `sl` | `I agree to the :terms_of_service and :privacy_policy` | `:terms_ of_service` | stray space |
| `sq` | same | `:privacy_politika` | token translated |
| `sr_Cyrl` | same | `:terms_оф_сервице` `:privacy_полици` | token transliterated |
| `sr_Latn` | same | `:terms_OF_SERVICE` `:privacy_POLICI` | case |
| `sr_Latn_ME` | same | `:terms_OF_SERVICE` `:privacy_POLICI` | case |

**Only the placeholder token is altered.** Not one word of translation is added,
removed or reworded, so none of this needs a native speaker to review — the diff
is nine single-token repairs.

One choice worth stating: the Serbian values keep the placeholder in lower case
rather than matching the surrounding capitals. `:TERMS_OF_SERVICE` is a form
`strtr()` does accept, but it upper-cases the substituted value, and in Jetstream
that value is an anchor tag with a URL inside it.

These nine are the subset that could be fixed mechanically. The same breakage
affects multi-word placeholders in about twenty more locales, and a much larger
problem sits behind them — in `as`, `sa`, `mni_Mtei` and `ckb` every placeholder
was replaced by a literal number. Both are described in the issue rather than
guessed at here.

### DIFF
--- a/locales/hu/json.json
+++ b/locales/hu/json.json
@@ -384,7 +384,7 @@
     "Liechtenstein": "Liechtenstein",
     "Light": "Fény",
     "Lithuania": "Litvánia",
-    "Load :perPage More": "Betöltése :perpage több",
+    "Load :perPage More": "Betöltése :perPage több",
     "Loading": "Betöltés",
     "Location": "Elhelyezkedés",
     "Log in": "Bejelentkezés",
--- a/locales/ko/json.json
+++ b/locales/ko/json.json
@@ -411,7 +411,7 @@
     "Manage API Tokens": "API 토큰 관리",
     "Manage Role": "역할 관리",
     "Manage Team": "팀 관리",
-    "Managing billing for :billableName": ":BilableName 에 대한 결제 관리",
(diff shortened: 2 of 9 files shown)


andrey-helldar (approved):
Thank you!

PR #8384, opened 2026-09-12
[tg]: Rewrite the inline validation messages

ismoil77:
Follow-up to #8381, this time for `locales/tg/php-inline.json`.

This file was in a worse state than `php.json`. Out of 138 keys, **38 were still in English** and most of the remaining ones were damaged in one way or another.

### Russian words left inside a Tajik sentence

| key | before |
| --- | --- |
| `active_url` | Ин **недопустимый URL-адрес**. |
| `string` | Ин аст, ки бояд **струна**. |
| `not_regex`, `regex`, `url` | Ин формат **недопустим**. |
| `exists`, `in`, `not_in` | Интихоб арзиши як **недопустимым**. |
| `json` | Ин бояд **допустимая** сатри JSON. |
| `same` | Аҳамияти ин соҳаи бояд **совпадать** … |

### Sentences that break off or lost their meaning

| key | before | what it says now |
| --- | --- | --- |
| `unique` | Ин аллакай анҷом. | "this is already finished" |
| `in_array` | Ин муҳим нест, ки дар :other. | "this is not important" |
| `file` | Маълумоте, ки бояд бошад, ки файли матнӣ намебошад. | "…is not a text file" |
| `prohibited_if_declined` | Ин майдон ҳангоми :other рад карда мешавад. | the "prohibited" part is gone |
| `filled` | Ин соҳа бояд дорои аҳамияти. | cut off |
| `alpha` | Ин соҳа метавонад дар бар гирад, ки танҳо ҳарфҳои. | cut off |
| `image` | Ин бояд тасвир. | no verb |
| `distinct` | Дар ин соҳа дорад, ки **бд**, аҳамияти. | leftover characters |
| `after`, `before`, `date_equals` | … пас аз **соли** :date. | stray "соли" (year) inside a date message |
| `starts_with` | Ин бояд **бо оғоз мекунанд, ки бо** яке аз … | word order |
| `doesnt_start_with`, `doesnt_end_with` | … **метавонад** … нашавад | "may not" instead of "must not" |

### How the new wording was chosen

The text comes from `locales/tg/php.json` (as fixed in #8381), rewritten into the impersonal form this file uses, so the two files agree with each other instead of drifting apart.

The comparison groups now follow one pattern per subject:

```
gt.numeric   Ин қиммат бояд аз :value зиёд бошад.
gt.string    Миқдори аломатҳо бояд аз :value зиёд бошад.
gt.file      Ҳаҷми файл бояд аз :value килобайт зиёд бошад.
[...]

PR #8381, opened 2026-09-12
[tg]: Fix Russian leftovers and broken grammar in validation messages

ismoil77:
Several messages in the Tajik validation file were left half-translated from Russian, or lost their grammar along the way, so they read as broken text to a Tajik speaker.

### Russian left inside the sentence

| key | before |
| --- | --- |
| `password` | Рамз **неверный**. |
| `not_regex` | Формати :attribute **недопустим**. |
| `attached` | Ин :attribute аллакай **прикреплен**. |
| `decimal` | :Attribute бояд :decimal адад **разряд** дошта бошад. |
| `ends_with` | **Рақами** :attribute бояд **заканчиваться** яке аз зерин: :values. |

### Wrong meaning, not just wording

| key | before | problem |
| --- | --- | --- |
| `doesnt_end_with` | :Attribute **метавонад** … хотима **наёбад** | says "may not end with" instead of "must not" |
| `doesnt_start_with` | :Attribute **метавонад** … оғоз **нашавад** | same |
| `lte.string` | :Attribute бояд камтар аз :value аломат бошад. | identical to `lt.string`, "or equal" is lost |
| `uuid` | :Attribute бояд **дархост эътибор дорад** UUID. | words don't form a sentence |
| `url` | Формати :attribute нодуруст мебошад. | that is the `regex` message, not "must be a valid URL" |
| `prohibited` | Майдони :attribute **сол** манъ аст. | stray word ("сол" means "year") |
| `missing` | Майдони :attribute бояд **нест**. | ungrammatical |
| `lowercase` | :Attribute бояд **хурд** бошад. | reads "must be small", nothing about letters |
| `date_equals` | :Attribute бояд санаи**,** баробар :date. | stray comma, sentence breaks off |
| `multiple_of` | Шумораи :attribute бояд **multiples** :value | English word, no final stop |
| `throttled` | … пеш аз он ки такрор **мекунед**. | wrong verb form |
| `starts_with` | :Attribute бояд **оғоз бо** яке аз арзишҳои зерин | word order |
| `array_keys`, `base64` | *(still English)* | not translated yet |

`uppercase` and `missing_unless` are adjusted too, so that they stay consistent with their pairs `lowercase` and `missing`.

### How the wording was chosen

- It follows the terms already used elsewhere in this file: `Қиммати :attribute` (37 keys), `майдон`, `гузарвожа`, `аломат`, `мавҷуд`, `манъ`.
[...]

PR #8385, opened 2026-09-12
[tg]: Fix machine-translated strings in json.json

ismoil77:
I am a native Tajik speaker. This started as a small fix for Russian words left in the file, but when I went through the whole file I found the locale is mostly raw machine translation, so I fixed all of it. 419 strings out of 871.

What was wrong:

**Country names translated as common words.** `Dominica` was «Якшанбе» (Sunday), `El Salvador` was «Наҷотдиҳанда» (the Savior), `Mali` was «Хурд» (small), `Turkey` was «Мурғи марҷон» (the bird), `Togo` was «Аз он» (from it), and `Suriname` was «Узбакистон» (Uzbekistan).

**Russian text.** `Address` → «Адрес», `Whoops` → «Упс», `Page Expired` → «Page Просрочена», `Service Unavailable` → «Хизматрасонии Недоступна», `Western Sahara` → «Западная Сахара», `United Kingdom` → «Объединенное Салтанат». The heading `Two Factor Authentication` was «Двухфакторная, шумо бояд ворид шавед», which is a piece of a Russian sentence.

**English text.** `View` → «Watch», `Administrator` → «Admin», `Force Delete` → «Маҷбуран uninstall», `Show Content` → «Нишон content», `Delete File` → «Чӣ тавр ба хориҷ image».

**Broken grammar.** Every `Delete *` string started with «Чӣ тавр ба хориҷ», which means "how to exclude". Many strings ended with a dangling izafat («Амали», «Навъи», «Нақши», «Захираҳои») — in Tajik that ending promises a following word, so the label reads as unfinished. `Regenerate Recovery Codes` was «Регенератсияи кислота...» — "regeneration of acid".

Also `You have enabled two factor authentication.` and `You have not enabled two factor authentication.` had the same translation, so the user could not tell the two states apart. And `Load :perPage More` had the placeholder broken into `:per` + a word, so it would never be replaced.

How I checked myself:

- terminology is now consistent through the file: password — гузарвожа, delete — нест кардан, resource — захира, user — корбар, field — майдон, two factor authentication — тасдиқи дуомила;
- `navigation` follows the Tajik MediaWiki interface, which uses «новбарӣ» (so `Pagination Navigation` is «Новбарии саҳифаҳо», with the izafat on the right word);
- country names follow their Tajik labels on Wikidata / Tajik Wikipedia, not my own transliteration;
- where I was not sure about a word I compared with Persian, which is the closest language, and checked the word really exists in Tajik Wikipedia.

Key count, key order and all placeholders are unchanged — only the values.

This is the same work as #8381 (php.json) and #8384 (php-inline.json) but for json.json.


andrey-helldar (approved):
Ташаккур

PR #8408, opened 2026-09-20
[es]: Update Spanish Translations

luisprmat:
Translate new keys (encrypted environment file) and fix some countries that do have a Spanish translation.


luisprmat:
While running some tests, I realized that some of these translations aren't working.

Several keys were extracted (by `status-generator` parser) from the framework's source code because they matched the expression `'/\b(fail)\(\r*\s*([\'"])/'`.

But this expression appears in commands `$this->fail('...')` where there is no support for translation. For example [`A decryption key is required.`](https://github.com/laravel/framework/blob/1d9727160ad440d1ccf2fdcbb7e4f4f36efae212/src/Illuminate/Foundation/Console/EnvironmentDecryptCommand.php#L72) is not translatable but it's appears in [`lang/source/framework/13.x/framework.json`](https://github.com/Laravel-Lang/lang/blob/main/source/framework/13.x/framework.json#L2).

This would only make sense if used within a validation rule.
```php
<?php
 
namespace App\Rules;
 
use Closure;
use Illuminate\Contracts\Validation\ValidationRule;
 
class Uppercase implements ValidationRule
{
    /**
     * Run the validation rule.
     */
    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        if (strtoupper($value) !== $value) {
            $fail('The :attribute must be uppercase.')->translate();
        }
    }
}
```
but we must chain the method `translate` in the call to the `$fail` method in order to translate it, since this method is an instance of class `\Illuminate\Translation\PotentiallyTranslatedString`.

andrey-helldar (approved):
Muchas gracias

